PR #8, opened 2020-09-19
Messaging refactor

rhlsthrm:


modules/protocol/src/sync.ts
```diff
@@ -44,77 +38,35 @@ export async function outbound(
...
-      await delay(3_000);
-    }
+  logger.info({ method: "outbound", to: update.toIdentifier, type: update.type }, "Sending protocol message");
+  const result = await messagingService.sendProtocolMessage(update, storedChannel?.latestUpdate);
 
-    return Result.fail(new ChannelUpdateError(ChannelUpdateError.reasons.MessageFailed, updateToSend, storedChannel));
-  };
+  if (result.isError) {
+    logger.error({ method: "outbound", error: result.getError()! }, "Error receiving response!");
+    return Result.fail(result.getError()!);
```
LayneHaber, line 46: can't you just `return result`

rhlsthrm: No, because `result` has a different type of value from the called function.

LayneHaber (approved):

PR #10, opened 2020-09-22
Add CAIP support

pedrouid:
- [x] add CAIP utils
- [ ] replace assetId on engine
- [ ] replace assetId with assetAddress on protocol


LayneHaber:
closing for now

PR #44, opened 2020-10-01
Init json config

bohendo:


rhlsthrm:
Added full Typebox schema validation to the config. Also added router config. Notes for @bohendo:

- Should we separate out node and router schemas in the config file, then use `jq` to pass in only the relevant ones (+ common config)?
- Router requires a `nodeUrl` config item. This will of course change per deployment, so we need a way to deal with this.

PR #109, opened 2020-10-21
105 in memory lock service

### LINKED ISSUE #105: [server-node] Implement in-memory lock service

rhlsthrm:
## The Problem

- Need to operate server node in single-container mode.

## The Solution

Fixes #105


ArjunBhuptani (approved):

LayneHaber (approved):
Wow that was quite the narrow race condition wasn't it

LayneHaber (changes requested):
Do docs need to be updated now that we aren't always running redis?

rhlsthrm:
> Do docs need to be updated now that we aren't always running redis?

Yes, this is correct but ops are not updated yet, that will be done in a separate PR.

PR #114, opened 2020-10-24
Contract tests

LayneHaber:
## Description
Includes the majority of contract tests, and associated bug  fixes:
- `nonReentrant` modifiers added
- changes to getters and function names for consistency
- interface definition extends `IAssetTransfer` for `ICMCAccountant`
- transfers use a *specific* cancelling resolver rather than cancelling if an incorrect resolver is used

There are still some outstanding issues/tests to be addressed:
1. `ChannelFactory.getChannelAddress()` failing #115 
2. `CMCAdjudicator` tests are not implemented #116
3. `LibIterableMapping.sol` tests are not implemented #117
4. transfer of USDT like assets failing #118


LayneHaber:
NOTE: This should **not** be published until after the contracts are deployed

PR #123, opened 2020-10-26
115 fix channel address computation

### LINKED ISSUE #115: [contract] ChannelFactory.getChannelAddress
**Describe the bug**
[This](https://github.com/connext/vector/blob/d16afd30a01f70e6a0cfc000c1ce016b1cc67de0/modules/contracts/src.ts/tests/channelFactory.spec.ts#L56) test is failing. Most likely because we are not left-padding the address to 256 bits.

HeikoFisch:
Closes #115.


bohendo (approved):
Looks great

PR #126, opened 2020-10-26
116 adjudicator changes

### LINKED ISSUE #116: [contracts] CMCAdjudicator finalized + tests
**Describe the bug**
Adjudicator tests have not been implemented. There are also some changes needed:

1. We should *not* support disputing at the same nonce
2. Anyone can initiate and complete a dispute

HeikoFisch:
Closes #116.


LayneHaber:
NOTE:

Currently the adjudicator uses a `msg.sender` check in the `defundTransfer` function. Instead, I think this should be a signature check. Waiting on final word from @HeikoFisch but this could be an additional change needed in the adjudicator we should be aware of

HeikoFisch:
> Currently the adjudicator uses a `msg.sender` check in the `defundTransfer` function. Instead, I think this should be a signature check.

From a security perspective, the more flexible signature-based approach should be fine. However, like @ArjunBhuptani, I don't see a convincing case where we'd actually make use of that flexibility; as soon as the responder knows the resolver data, they'd probably rather just use it than save it for later. We can still change it "just in case" or because it fits well into the general adjudicator pattern, where we work a lot with signatures and this would be the only place with `msg,sender`-based authentication; on the other hand, the simple `msg.sender` check is more gas-efficient and requires less code.
I don't have a strong preference here; if you're more comfortable with the signature-based approach, let me know and we can work out the details.

PR #128, opened 2020-10-27
ParameterizedPayments + InsurancePayment transferDefs

ditchfieldcaleb:
ParameterizedPayment + Arachnid's string utils library: [https://github.com/Arachnid/solidity-stringutils](https://github.com/Arachnid/solidity-stringutils)

### DIFF
--- a/modules/contracts/src.sol/lib/strings.sol
+++ b/modules/contracts/src.sol/lib/strings.sol
@@ -1,0 +1,716 @@
+/*
+ * @title String & slice utility library for Solidity contracts.
+ * @author Nick Johnson <arachnid@notdot.net>
+ *
+ * @dev Functionality in this library is largely implemented using an
+ *      abstraction called a 'slice'. A slice represents a part of a string -
+ *      anything from the entire string to a single character, or even no
+ *      characters at all (a 0-length slice). Since a slice only has to specify
+ *      an offset and a length, copying and manipulating slices is a lot less
+ *      expensive than copying and manipulating the strings they reference.
+ *
+ *      To further reduce gas costs, most functions on slice that need to return
+ *      a slice modify the original one instead of allocating a new one; for
+ *      instance, `s.split(".")` will return the text up to the first '.',
+ *      modifying s to only contain the remainder of the string after the '.'.
+ *      In situations where you do not want to modify the original slice, you
+ *      can make a copy first with `.copy()`, for example:
+ *      `s.copy().split(".")`. Try and avoid using this idiom in loops; since
+ *      Solidity has no memory management, it will result in allocating many
+ *      short-lived slices that are later discarded.
+ *
+ *      Functions that return two slices come in two versions: a non-allocating
+ *      version that takes the second slice as an argument, modifying it in
+ *      place, and an allocating version that allocates and returns the second
+ *      slice; see `nextRune` for example.
+ *
+ *      Functions that have to copy string data will return strings rather than
+ *      slices; these can be cast back to slices for further processing if
+ *      required.
+ *
+ *      For convenience, some functions are provided with non-modifying
+ *      variants that create a new slice and return both; for instance,
+ *      `s.splitNew('.')` leaves s unmodified, and returns two values
+ *      corresponding to the left and right parts of the string.
+ */
+
+pragma solidity ^0.4.14;
+
+library strings {
+    struct slice {
+        uint _len;
+        uint _ptr;
+    }
+
+    function memcpy(uint dest, uint src, uint len) private pure {
+        // Copy word-length chunks while possible
+        for(; len >= 32; len -= 32) {
+            assembly {
+                mstore(dest, mload(src))
+            }
+            dest += 32;
+            src += 32;
+        }
+
+        // Copy remaining bytes
+        uint mask = 256 ** (32 - len) - 1;
+        assembly {
+            let srcpart := and(mload(src), not(mask))
+            let destpart := and(mload(dest), mask)
+            mstore(dest, or(destpart, srcpart))
+        }
+    }
+
+    /*
+     * @dev Returns a slice containing the entire string.
+     * @param self The string to make a slice from.
+     * @return A newly allocated slice containing the entire string.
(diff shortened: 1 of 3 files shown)


modules/contracts/src.sol/lib/strings.sol
```diff
@@ -0,0 +1,716 @@
...
+ *      slices; these can be cast back to slices for further processing if
+ *      required.
+ *
+ *      For convenience, some functions are provided with non-modifying
+ *      variants that create a new slice and return both; for instance,
+ *      `s.splitNew('.')` leaves s unmodified, and returns two values
+ *      corresponding to the left and right parts of the string.
+ */
+
+pragma solidity ^0.4.14;
```
LayneHaber, line 37: We use solidity `0.7.1`, not sure if it's possible to upgrade this definition but it may act up in some of our other build scripts (will have to try manually)

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
```diff
@@ -0,0 +1,139 @@
...
+  string RateEncoding = "tuple(uint256 deltaAmount, uint256 deltaTime)";
+  string StateEncoding = 
+    "tuple(address receiver, uint256 start, uint256 expiration, bytes32 UUID, "
+    .concat(RateEncoding)
+    .concat(")");
+  string ResolverDataEncoding = "tuple(bytes32 UUID, uint256 paymentAmountTaken)";
+  string ResolverEncoding = 
+    "tuple("
+    .concat(ResolverDataEncoding)
+    .concat(", bytes payeeSignature)");
```
LayneHaber, line 61: It would be a good idea to make sure these work (we've been burned before). It's not too bad to do it in the node repl, but would probably have to be done manually

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
```diff
@@ -0,0 +1,139 @@
...
+  ) external override view returns (Balance memory) {
+    // Unpack balance, state, resolver, & resolverdata
+    Balance memory balance = abi.decode(encodedBalance, (Balance));
+    TransferState memory state = abi.decode(encodedState, (TransferState));
+    TransferResolver memory resolver = abi.decode(encodedResolver, (TransferResolver));
+    ResolverData memory data = resolver.data;
+
+    // Signature check
+    bytes32 hashedData = keccak256(abi.encode(data));
+    require(state.receiver == LibChannelCrypto.verifyChannelMessage(hashedData, resolver.payeeSignature));
```
LayneHaber, line 119: Would probably be a good idea to add some revert messages around these requires

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
```diff
@@ -0,0 +1,139 @@
...
+    require(block.timestamp < state.expiration);
+
+    // Rate should not be exceeded; multiply by large number to avoid precision errors
+    uint256 timeElapsed = block.timestamp.sub(state.start);
+    uint256 averageRate = data.paymentAmountTaken.div(timeElapsed.multiply(2**64));
+    uint256 allowedRate = state.rate.deltaAmount.div(state.rate.deltaTime.multiply(2**64));
+    require(averageRate <= allowedRate);
+
+    // Transfer the payment amount
+    balance.amount[0] = balance.amount[0].sub(data.paymentAmountTaken);
```
LayneHaber, line 134: should there be a require that the `paymentAmountTaken` is not larger than the original amount in the transfer?

i know technically safe math would save you here, but i do love my sanity checks

LayneHaber (commented):
Also, we are doing some internal debate about the best way to include transfer definitions in the code, so this PR may be open while we settle on the best flow for that. We can add the transfer def to a rinkeby registry to make sure you're not blocked

modules/contracts/src.sol/lib/strings.sol
ditchfieldcaleb: Resolved per conversation in Discord.

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
ditchfieldcaleb: @LayneHaber can you give an overview of how I'd test these?

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
ditchfieldcaleb: @LayneHaber wouldn't that fail when the transfer tries to resolve with a higher balance than in the transfer itself?

ditchfieldcaleb:
@LayneHaber @ArjunBhuptani I've gone ahead and combined both payment types into this PR since they're very related and essentially follow the same code flow. Would love to hop on a call tomorrow for a thorough review if either of you have time!

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
ditchfieldcaleb: Added in 188bfb124fb4287ef7d27700d9d2b1f410fa0e04

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
ditchfieldcaleb: Confirmed with @ArjunBhuptani , but added in just to be safe!

modules/contracts/src.sol/transferDefinitions/ParameterizedPayments.sol
ditchfieldcaleb: Testing locally, will update.

LayneHaber:
Okay, we've created a new repository to host all of the transfer definitions [here](https://github.com/connext/transfers).

I'm going to close this PR -- it can be resubmitted to the new repo (which should be properly configured for tests)

PR #140, opened 2020-11-02
118 121 asset transfers

### LINKED ISSUE #118: [contracts] USDT-like asset transfer
**Describe the bug**
The transfer of USDT-like assets currently results in a double-spend, and we should be more restrictive about what we consider a failed transfer.

We should always bias towards locked funds over double spends.

### LINKED ISSUE #121: [contracts] Alice should be able to deposit USDT-like tokens
Alice should be able to deposit missing-return-value-bug tokens, like USDT. That's not implemented yet.

HeikoFisch:
Closes #118.
Closes #121.


modules/contracts/src.sol/AssetTransfer.sol
```diff
@@ -72,15 +66,17 @@ contract AssetTransfer is CMCCore, IAssetTransfer {
   }
 
   function _getOwnERC20Balance(address assetId) external view onlySelf returns (uint256) {
-    return IERC20(assetId).balanceOf(address(this));
+    return IERC20(assetId).balanceOf{gas: ERC20_BALANCEOF_GAS_LIMIT}(address(this));
```
LayneHaber, line 69: it seems in the majority of this PR we've replaced the calls to `IERC20` with calls to `LibERC20`. these changes are really only needed for transfers, but we may want to move this call in there as well for consistency

modules/contracts/src.sol/CMCWithdraw.sol
```diff
@@ -15,26 +16,30 @@ contract CMCWithdraw is CMCCore, AssetTransfer, ICMCWithdraw {
...
     isExecuted[withdrawHash] = true;
 
     // Validate signatures
     require(withdrawHash.checkSignature(aliceSignature, alice), "CMCWithdraw: Invalid alice signature");
     require(withdrawHash.checkSignature(bobSignature, bob), "CMCWithdraw: Invalid bob signature");
 
+    // Determine actually transferable amount
+    uint256 balance = LibAsset.getOwnBalance(assetId);
+    uint256 amount = LibUtils.min(maxAmount, balance);
+
```
LayneHaber, line 42: nice catch

modules/contracts/src.sol/AssetTransfer.sol
HeikoFisch: Sure, we can do that. Right now, the `LibERC20` only wraps the calls that behave differently in standard-compliant and missing-return-value-bug tokens, so it is consistent in a way. But I get your drift and am also fine with the suggested alternative.

HeikoFisch: Actually, on second thought, I'd prefer to leave it as it is, because we cannot restrict the amount of gas passed on for an _internal_ call, so we'd have to pass it along as a parameter and then have a special `balanceOf` function in the `LibERC20` that takes a gas parameter. That would be a bit ugly, and in the end, the overhead doesn't seem worth it. You're ok with that, @LayneHaber?

LayneHaber: Yeah sg, will approve + merge 👍

LayneHaber (approved):

PR #142, opened 2020-11-03
Transfer in BrowserNode

sanchaymittal:
## The Problem

Transfer for browser node is not working.

## The Solution

- function getPublicKeyFromPublicIdentifier allows getting the public key from the recipient public identifier
- then function encrypt allows encrypting preImage & publicKey to get encryptedPreImage.
- and putting it into the meta property in the conditionalTransfer call so that when the event is received by the receiver it will have access to that meta.


modules/test-ui/src/App.tsx
```diff
@@ -134,6 +157,14 @@ function App() {
 
   const transfer = async (assetId: string, amount: string, recipient: string, preImage: string) => {
     setTransferLoading(true);
+
+    const submittedMeta: { encryptedPreImage?: any } = {};
```
rhlsthrm, line 161: Can we make this something that's not `any`? It should be `string` right?

modules/test-ui/src/utils/crypto.ts
```diff
@@ -0,0 +1,32 @@
+import { HexString, PublicKey, PrivateKey } from "@connext/vector-types";
```
rhlsthrm, line 1: Let's move this stuff into the `@connext/vector-utils` package instead of in the test-ui.

modules/test-ui/src/utils/hexStrings.ts
```diff
@@ -0,0 +1,21 @@
+import { utils } from "ethers";
```
rhlsthrm, line 1: Same as above.

modules/test-ui/src/utils/identifiers.ts
```diff
@@ -0,0 +1,10 @@
+import { PublicIdentifier } from "@connext/vector-types";
+import bs58check from "bs58check";
+import { bufferToHex, decompress } from "eccrypto-js";
+
+export const INDRA_PUB_ID_PREFIX = "indra";
+
+export const getPublicKeyFromPublicIdentifier = (publicIdentifier: PublicIdentifier) =>
```
rhlsthrm, line 7: This function should already be in `@connext/vector-utils`, just go ahead and import it from there.

rhlsthrm (changes requested):
Minor fixes requested.

modules/test-ui/src/utils/crypto.ts
sanchaymittal: Ahh didn't notice we already had utils. Done

rhlsthrm (approved):

PR #143, opened 2020-11-03
32 generalized withdrawals

### LINKED ISSUE #32: [contracts] Move back to `execWithdraw` pattern where arbitrary withdraw txs can be executed
We *should not* restrict this at the contract level

HeikoFisch:
Closes #32.


modules/contracts/src.sol/CMCWithdraw.sol
```diff
@@ -3,57 +3,98 @@ pragma solidity ^0.7.1;
...
+    require(!isExecuted[wdHash], "CMCWithdraw: Transaction has already been executed");
+    isExecuted[wdHash] = true;
 
     // Determine actually transferable amount
-    uint256 balance = LibAsset.getOwnBalance(assetId);
-    uint256 amount = LibUtils.min(maxAmount, balance);
+    uint256 balance = LibAsset.getOwnBalance(wd.assetId);
+    uint256 amount = LibUtils.min(wd.amount, balance);
+
+    // TODO: Should we revert if amount is 0?
```
LayneHaber, line 60: There could be some cases where we want to execute the additional call, even with a 0 balance (can't think of any, but I'm sure someone else could) so we probably don't want to revert here unless the `amount` and `callTo` are empty

HeikoFisch: I agree. In fact, I don't even think it's necessary to revert if both `amount` and `callTo` are zero; we could just "let it run" in this case too (which is what I implemented, but I felt I shouldn't "silently" decide this on my own, so I added the comment).
Do you think reverting would be better if both are zero? (One thing that might be worth considering in this context are retries of failed transactions.)

LayneHaber: We only retry on known errors (i.e. nonce errors), so I don't think we have to worry too much about the retries on reverts.

I think reverting iff both are zero is a good call, since its very likely that's not what you wanted to do anyways

LayneHaber (approved):

PR #147, opened 2020-11-05
Browser node sdk

rhlsthrm:


modules/browser-node/src/channelProvider.ts
```diff
@@ -0,0 +1,280 @@
...
+export interface IframeOptions {
+  id: string;
+  src: string;
+}
+
+export function payloadId(): number {
+  const date = new Date().getTime() * Math.pow(10, 3);
+  const extra = Math.floor(Math.random() * Math.pow(10, 3));
+  return date + extra;
+}
```
LayneHaber, line 30: Is this something we have to do in the server node as well?

modules/browser-node/src/channelProvider.ts
```diff
@@ -0,0 +1,280 @@
...
+    event: T,
+    callback: (payload: EngineEventMap[T]) => void | Promise<void>,
+    filter?: (payload: EngineEventMap[T]) => boolean,
+  ): Promise<void>;
+  once<T extends EngineEvent>(
+    event: T,
+    callback: (payload: EngineEventMap[T]) => void | Promise<void>,
+    filter?: (payload: EngineEventMap[T]) => boolean,
+  ): Promise<void>;
+}
```
LayneHaber, line 56: This seems like something that should go into the `types` module

modules/iframe-app/config-node.json
```diff
@@ -0,0 +1,29 @@
+{
```
LayneHaber, line 1: This should probably be `.gitignored` like our other config files

modules/iframe-app/local.config-node.json
```diff
@@ -0,0 +1,29 @@
+{
```
LayneHaber, line 1: This should go with our other default config files in `ops/config`

modules/iframe-app/prod.config-node.json
```diff
@@ -0,0 +1,18 @@
...
+      "hashlockTransferAddress": "0xd69ABEf7eF08957F47d5B060B63D0b71Ee8d4621",
+      "withdrawAddress": "0xc5E4115B8fB081c828Fe65763e1038152FAA3213",
+      "transferRegistryAddress": "0x268ee7145a1c717d36dAfcF402b51ac4BAdD6388"
+    }
+  },
+  "chainProviders": {
+    "4": "https://rinkeby.infura.io/v3"
+  },
+  "port": 3333
+}
```
LayneHaber, line 18: Should also go in `ops/config`

modules/server-node/ops/entry.sh
```diff
@@ -30,6 +30,7 @@ else
   touch "$sqlite_file"
   export VECTOR_DATABASE_URL="sqlite://$sqlite_file"
 fi
+echo "VECTOR_DATABASE_URL: $VECTOR_DATABASE_URL"
```
LayneHaber, line 33: Can take this out

modules/test-ui/config-node.json
```diff
@@ -25,5 +25,5 @@
     "1337": "http://localhost:8545",
     "1338": "http://localhost:8546"
   },
-  "port": 3333
+  "port": 3000
```
LayneHaber, line 28: should either be `gitignored` or put in the config directory

modules/test-ui/src/App.tsx
```diff
@@ -1,69 +1,62 @@
...
-    init();
+    effect();
   }, []);
 
-  const connectNode = async (mnemonic: string) => {
-    console.log("creating node with config", config);
+  const connectNode = async (iframeSrc: string, entropy: string): Promise<BrowserNode> => {
+    if (!iframeSrc) {
+      iframeSrc = "http://localhost:3030";
+    }
```
LayneHaber, line 53: Shouldn't this be derived from the config?

modules/test-ui/src/App.tsx
```diff
@@ -77,21 +70,27 @@ function App() {
...
       }
       setNode(client);
-      localStorage.setItem("mnemonic", mnemonic);
-      setMnemonic(mnemonic);
       client.on(EngineEvents.DEPOSIT_RECONCILED, async data => {
         console.log("Received EngineEvents.DEPOSIT_RECONCILED: ", data);
         await updateChannel(client, data.channelAddress);
       });
+      // TODO: this is required bc the event handlers are keyed on Date.now()
+      await delay(10);
```
LayneHaber, line 78: Is this still required with the modified `+ extra`?

modules/test-ui/src/App.tsx
```diff
@@ -77,21 +70,27 @@ function App() {
...
-        console.log(data.transfer.meta.encryptedPreImage, wallet.privateKey);
-        const decryptedPreImage = await decrypt(data.transfer.meta.encryptedPreImage, wallet.privateKey);
-        console.log(decryptedPreImage);
+        console.log(data.transfer.meta.encryptedPreImage);
+        const rpc: EngineParams.RpcRequest = {
+          id: Date.now(),
+          jsonrpc: "2.0",
+          method: "chan_decrypt",
+          params: data.transfer.meta.encryptedPreImage,
+        };
```
LayneHaber, line 91: Should use the utils for this

modules/utils/src/rpc.ts
```diff
@@ -1,7 +1,7 @@
 import { ChannelRpcMethod, ChannelRpcMethodsPayloadMap, EngineParams } from "@connext/vector-types";
 
 export const constructRpcRequest = <T extends ChannelRpcMethod>(
-  method: T,
+  method: any,
```
LayneHaber, line 4: Should be typed as `T` right?

LayneHaber (changes requested):

modules/browser-node/src/channelProvider.ts
rhlsthrm: Nope this is only for the iframe <> parent communication through postMessage.

modules/browser-node/src/channelProvider.ts
rhlsthrm: It's only relevant to the browser-node. I chatted about this with @bohendo too, we agree that only shared types should go into the types module.

modules/iframe-app/config-node.json
rhlsthrm: Definitely! Fixed.

modules/iframe-app/local.config-node.json
rhlsthrm: You're right about this. The whole config strategy has changed since now this is all in the iframe. The config for the iframe just needs to be the default node config now. Fixed this to use the same config.

modules/iframe-app/prod.config-node.json
rhlsthrm: Fixed.

modules/test-ui/config-node.json
rhlsthrm: Done.

modules/test-ui/src/App.tsx
rhlsthrm: The app (test UI is basically an example of a "dapp" that integrates Connext browser node) does not know anything about the iframe's config. It's basically a black box.

modules/test-ui/src/App.tsx
rhlsthrm: What's the `+ extra`?

modules/test-ui/src/App.tsx
rhlsthrm: Fine, lol. Fixed.

modules/utils/src/rpc.ts
rhlsthrm: I could swear I tried this last time and it didn't work that's why I had to change it but it seemed to work now.

LayneHaber (approved):
Some lingering console.logs but looks great!

PR #148, opened 2020-11-06
Browser node automation test

sanchaymittal:
- Cypress config files
- Router Request
- Window web page
- Node profile creation
- channel setup spec

Ref: https://www.notion.so/connext/Browser-Node-Automation-Test-d359073e32f041c0b12497c10990a1d8


rhlsthrm (approved):
Would like to see if we can clean up some of the HTML selectors but other than that it looks great.

PR #150, opened 2020-11-08
vector contracts deployed on maticmumbai

Aveesh52:
Deployed vector contracts on maticmumbai,now address-book.json has the address of deployed contracts indexed by maticmumbai's chainId so that they're readily available for everyone else to use

PR #151, opened 2020-11-08
maticmumbai added

Aveesh52:
Deployed vector contracts on maticmumbai,now address-book.json has the address of deployed contracts indexed by maticmumbai's chainId so that they're readily available for everyone else to use

Happy case tests are passing with Matic ,having some issue with withdrawals for carol which will be fixed


LayneHaber (changes requested):
It looks like the only substantial changes here are the ones in the address book, which should be included in #150 -- can you explain the difference here?

PR #155, opened 2020-11-11
Arbitrum

LayneHaber:
## The Problem
Our stuff didn't run on arbitrum testnet

## The Solution
Now runs on arbitrum testnet 🎉


rhlsthrm (changes requested):
I'd much rather combine all migrations from a single PR into one migration. There's no need to have multiple migrations here. Other than that, looks good.

rhlsthrm (approved):

PR #157, opened 2020-11-12
34 speed up

### LINKED ISSUE #34: [ops] Speed up build
Look into using: https://github.com/a7ul/esbuild-node-tsc

bohendo:
## The Problem

closes #34 

## The Solution

issue 34 was already addresses as best as is feasible a little while ago
but this work is in the same category

### DIFF
--- a/modules/contracts/src.sol/transferDefinitions/HashlockTransfer.sol
+++ b/modules/contracts/src.sol/transferDefinitions/HashlockTransfer.sol
@@ -27,9 +27,9 @@ contract HashlockTransfer is TransferDefinition {
     TransferState memory state = abi.decode(encodedState, (TransferState));
     Balance memory balance = abi.decode(encodedBalance, (Balance));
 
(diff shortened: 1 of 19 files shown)


modules/contracts/src.sol/transferDefinitions/HashlockTransfer.sol
```diff
@@ -47,15 +47,15 @@ contract HashlockTransfer is TransferDefinition {
...
 
       // Update state
       balance.amount[1] = balance.amount[0];
       balance.amount[0] = 0;
     } else {
       // To cancel, the preImage must be empty (not simply incorrect)
-      require(resolver.preImage == bytes32(0), "Must provide empty hash to cancel payment");
-
+      // TODO: rm this assertion bc it will always succeed given the if statement that leads us here
+      require(resolver.preImage == bytes32(0), "HashlockTransfer: NONZERO_LOCKHASH");
```
LayneHaber, line 58: We could get here because it has expired and someone is trying to resolve as well

LayneHaber (approved):

PR #163, opened 2020-11-17
162 proxy reentrancy

### LINKED ISSUE #162: [contracts] Proxy reentrancy protection
## Expected Behavior
Blanket reentrancy protection for `ChannelMastercopy` via `Proxy`

## Current Behavior
Custom modifiers and getters for functions

## Possible Solution
Reposting from @HeikoFisch on discord:

> The usual reentrancy guard works just like a "mutex". Conceptually, there is a boolean variable entered that's initialized to false, and whenever a reentrancy-protected function is called, we first verify that entered is false (revert otherwise), then set it to true and, at the end of the call, set it back to false. (In reality, it's implemented a bit differently, in order to avoid the expensive flip from false to true each time we call such a function.)
Because we're using a proxy anyway (which just forwards every call to the mastercopy), we could just reentrancy-protect this forwarding function in the proxy, using the same approach as above. This would give us blanket reentrancy protection over the entire mastercopy; it's not possible anymore to forget the nonReentrant modifier on some function. So, in the mastercopy, we can throw them all out, as well as the self-written getters -- and still have everything reentrancy-protected, without boilerplate.
The downside is it increases the code size of the proxy and therefore the deployment costs. So maybe a better alternative is to have a separate contract with this functionality between the "real" proxy and the mastercopy: the proxy delegatecalls to that reentrancy guard contract, which again delegatecalls to the mastercopy. (There is only one instance of the reentrancy guard contract.) The disadvantage of this solution is an additional indirection, making each call a bit more expensive. It's also a bit difficult or at least cumbersome to extend the onlyViaProxy protection to this new contract, but that's more of a "convenience protection" anyway.

LayneHaber:
## The Problem
Fixes #162


LayneHaber:
Won't work, killing

PR #168, opened 2020-11-17
remove chainId args

### LINKED ISSUE #159: [contracts] use chainId opcode instead of user input
Uniswap uses a domain separator for the permit fn of their pool token ERC20 piece (to enable meta-transactions). In it, they're using the chainId() opcode to enforce that the chainId of the domainHash matches what the chain's opcode returns..

My instinct is to default to patterns that well-established projects are using & I'm worried that some dangerous funny business could arise from our reliance on user-supplied chainId values.

For reference, uniswap's usage of the chainId opcode: https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2ERC20.sol#L27

One requirement for this is that we need to stop using ganache as a testnet bc it's return value for the chainId opcode is hardcoded & not configurable. Deets: https://github.com/trufflesuite/ganache-cli/issues/709

bohendo:
This reverts commit 6ada4eb206ce4ed64cbaa96575f14318caba9c55.

## The Problem

closes #159 

## The Solution

For maximal compatibility even w chains that improperly implement the chainId opcode, the approach we're moving forward with is:
We store the value that was given in the constructor. If that value is 0, we return the chain id via opcode; otherwise the stored value

### DIFF
--- a/modules/contracts/src.sol/ChannelFactory.sol
+++ b/modules/contracts/src.sol/ChannelFactory.sol
@@ -12,9 +12,11 @@ import "./lib/MinimalProxyFactory.sol";
 /// @author Connext & Friends <hello@connext.network>
 contract ChannelFactory is IChannelFactory, MinimalProxyFactory {
     address private immutable mastercopy;
+    uint private immutable chainId;
 
-    constructor(address _mastercopy) {
+    constructor(address _mastercopy, uint _chainId) {
         mastercopy = _mastercopy;
+        chainId = _chainId;
     }
 
     ////////////////////////////////////////
@@ -23,6 +25,11 @@ contract ChannelFactory is IChannelFactory, MinimalProxyFactory {
     /// @dev Allows us to get the mastercopy that this factory will deploy channels against
     function getMastercopy() external override view returns(address) {
       return mastercopy;
+    }
+
+    /// @dev Allows us to get the chainId that this factory will use in the create2 salt
+    function getChainId() external override view returns(uint) {
+      return chainId;
     }
 
     function proxyCreationCode() external override view returns (bytes memory) {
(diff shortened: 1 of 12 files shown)


modules/contracts/src.sol/ChannelFactory.sol
```diff
@@ -44,7 +50,7 @@ contract ChannelFactory is IChannelFactory, MinimalProxyFactory {
     {
         return _calculateMinimalProxyDeploymentAddress(
             mastercopy,
-            generateSalt(alice, bob, chainId)
+            generateSalt(alice, bob)
```
LayneHaber, line 53: Doesn't the `chainId` still have to be included in the salt of the channel address?

LayneHaber: Ahh see it at the bottom, disregard

LayneHaber (approved):

LayneHaber:
Before merging would like @HeikoFisch to review, but looks good to me!

modules/contracts/src.sol/ChannelFactory.sol
```diff
@@ -25,6 +27,11 @@ contract ChannelFactory is IChannelFactory, MinimalProxyFactory {
       return mastercopy;
     }
 
+    /// @dev Allows us to get the chainId that this factory will use in the create2 salt
+    function getChainId() external override view returns(uint) {
+      return chainId;
+    }
+
```
HeikoFisch, line 34: Hmm, this is technically totally correct, but I worry that it might be misleading:
It's just a getter for the stored variable and, as such, follows the naming convention. However, the name of the function seems to suggest this returns the actual chain id we're using in the salt computation.
If a function that returns the value stored in the variable is actually useful (in tests, for example, to verify what gets stored), I'd suggest to deviate in this case from the naming convention and name the function `getStoredChainId`, just to avoid any potential confusion.
The `getChainId` function should, in my opinion, return the "real" chain id that is used in the salt computation.

HeikoFisch (approved):
Apart from the minor point in the comment, LGTM!

PR #169, opened 2020-11-17
165 asset array

### LINKED ISSUE #165: [contracts] Modify defundChannel to pass in an array of assetIds
This way, we allow for disputing a *newly* deposited assetId which may not yet be a part of the state.

For this to work, we also need to allow `defundChannel` to be called many times within the defund window. We can assume that after the channel has been defuned, any remaining funds there are dust and can be ignored -- i.e. for the purposes of "restarting" the channel post-dispute, this change to `defundChannel` *should not make any difference*.

LayneHaber:
## The Problem
Fixes #165 

## The Solution
Add in an `assetId[]` for the defund process, and track defunding nonces by asset id


modules/contracts/src.sol/CMCAdjudicator.sol
```diff
@@ -81,47 +87,72 @@ contract CMCAdjudicator is CMCCore, CMCAccountant, ICMCAdjudicator {
...
+        // The index was supposedly given -- we verify
+        index = indices[i];
+        require(assetId == ccs.assetIds[index], "CMCAdjudicator: INDEX_MISMATCH");
+      } else {
+        // we search through the assets in ccs
+        for (index = 0; index < ccs.assetIds.length; index++) {
+          if (assetId == ccs.assetIds[index]) {
+            break;
+          }
+        }
```
LayneHaber, line 126: If the asset id is not found in the `ccs.assetIds` there won't be any reverts here, if that's the case then im not sure what exactly the for loop is adding

modules/contracts/src.sol/CMCAdjudicator.sol
```diff
@@ -81,47 +87,72 @@ contract CMCAdjudicator is CMCCore, CMCAccountant, ICMCAdjudicator {
...
-      // Add unprocessed deposits to amounts
-      balance.amount[0] += _getTotalDepositsAlice(assetId) - ccs.processedDepositsA[i];
+      // Now, if `index`  is equal to the number of assets in ccs,
+      // then the current asset is not in ccs;
+      // otherwise, `index` is the index in ccs for the current asset
+
+      // Check the assets haven't already been defunded + update the
+      // defundNonce for that asset
+      {
+        // Open a new block to avoid "stack too deep" error
```
LayneHaber, line 136: This is very good to know

modules/contracts/src.sol/CMCAdjudicator.sol
LayneHaber: Ahh okay, I see. Correct if wrong:

The index is being assigned as the counter within the for loop, and if it does not exist in the channel state (i.e. a deposit of a new asset was sent to chain, but not reconciled into the channel), then it will be transferred out as an unprocessed deposit

modules/contracts/src.sol/CMCAdjudicator.sol
```diff
@@ -218,6 +249,21 @@ contract CMCAdjudicator is CMCCore, CMCAccountant, ICMCAdjudicator {
     );
   }
 
+  function getTotalDeposits(address assetId) internal returns (uint256 tdAlice, uint256 tdBob) {
+    tdAlice = _getTotalDepositsAlice(assetId);
+
+    // If the following call fails, we were unable to query the channel's balance;
+    // this probably means the token is totally fucked up.
```
LayneHaber, line 256: 😆

LayneHaber (commented):
LGTM -- though since some of the logic around the indices and asset ids is non-intuitive would like another set of eyes to make sure it lines up!

modules/contracts/src.sol/CMCAdjudicator.sol
HeikoFisch: Exactly.

modules/contracts/src.sol/CMCAdjudicator.sol
HeikoFisch: It does not always do the trick, but when it does, it's one of the less painful/code-mutilating countermeasures.

PR #183, opened 2020-11-21
Test remote node

LayneHaber:
## The Problem
Documents how to test a remote node from a local UI


ops/start-iframe-app.sh
```diff
@@ -14,6 +14,14 @@ then interactive=(--interactive --tty)
 else echo "Running in non-interactive mode"
 fi
 
+####################
+# Misc Config
+
+# Load the config with defaults if it does not exist
+if [[ ! -f "$root/browser.config.json" ]]
```
rhlsthrm, line 21: I think you can do -z instead of the ! -f. We do that elsewhere.

rhlsthrm (approved):

ops/start-iframe-app.sh
LayneHaber: ahh i copy-pasted from `start-trio`, so i'm gonna leave it

PR #198, opened 2020-11-25
195 register transfer

### LINKED ISSUE #195: [ops] Register transfer from transfers repo locally
Should be able to:

- deploy transfer to local vector setup
- register transfer against deployed registry
- work with new transfer definition

LayneHaber:
## The Problem

Fixes #195 

See the [transfer readme](https://github.com/connext/transfers) for information about how this should work with a local setup


LayneHaber:
Current problem: not able to get network ID from the eth provider at `http://localhost:8545`

rhlsthrm:
Fixed in transfer repo.

PR #213, opened 2020-11-30
fix approval bug

rhlsthrm:
## The Problem
Token approvals weren't working for Alice deposit.

## The Solution

Fixed chainReader.


modules/contracts/src.ts/services/ethService.ts
```diff
@@ -436,6 +438,7 @@ export class EthereumChainService extends EthereumChainReader implements IVector
     try {
       const response = await this.queue.add(async () => {
         const response = await txFn();
+        console.log("sendTxAndParseResponse =======> response: ", response);
```
LayneHaber, line 441: We can remove this

LayneHaber (approved):
Pending passing tests

PR #233, opened 2020-12-04
200 rm database

### LINKED ISSUE #200: [ops] rm ops/database in favor of a professionally hosted db
eg "prod-mode" should use an external db url. For testing or staging, the in-built sqlite db is sufficient.

Our current ops/database is kind of in an uncanny valley: it's homegrown so still vulnerable to bugs in production but still contains enough complicated back-up logic to complicate dev/staging deployments.

bohendo:
DO NOT MERGE until we're 100% sure that zero integrations are reliant on our homegrown db ops. Currently, XUD is the only known team that's using ops/database but we should be aware that there might be others.

There is no hurry at all. If it ends up being a major pain to migrate to a new db, this PR never needs to be merged.

## The Problem

closes #200 

## The Solution

`rm -rf ops/database`


bohendo:
WARNING: It just occurred to me that we're explicitly using sqlite in router-prod tests which is NOT correct because it means the router & node are using two separate databases. The tests are only passing because the prod router integration tests are so simple but do not take this as an indication that all is well.

PR #237, opened 2020-12-04
fix: initiator cannot be receiver on same chain.

sanchaymittal:
add: todo test in engine

## The Problem

@rhlsthrm: An initiator cannot be a receiver on the same chain
It seemed to work for one transfer leg and fails for another one. It was erroring at the DB level and causing things to be messed up

## The Solution

throw an error at `convertConditionalTransferParams()`

### DIFF
--- a/modules/engine/src/paramConverter.ts
+++ b/modules/engine/src/paramConverter.ts
@@ -37,7 +37,7 @@ export async function convertConditionalTransferParams(
   // TODO: Compare recipientChainId with signerChainId
   if (recipient === signer.publicIdentifier) {
     // If signer is also the receipient on same chain/network
-    throw new Error(`Invalid initiator cannot be receiver on same chain`);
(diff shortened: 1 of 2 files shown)


modules/engine/src/paramConverter.ts
```diff
@@ -34,6 +34,12 @@ export async function convertConditionalTransferParams(
   const recipientAssetId = params.recipientAssetId ?? params.assetId;
   const channelCounterparty = signer.address === channel.alice ? channel.bob : channel.alice;
 
+  // TODO: Compare recipientChainId with signerChainId
+  if (recipient === signer.publicIdentifier) {
+    // If signer is also the receipient on same chain/network
+    throw new Error(`Invalid initiator cannot be receiver on same chain`);
```
rhlsthrm, line 40: We should be returning a result type here not throwing error.

rhlsthrm (changes requested):

LayneHaber (approved):

PR #238, opened 2020-12-05
Pull receiver transfer definition from on chain

rhlsthrm:
## The Problem

Cross chain transfers weren't working IRL.

## The Solution

Pull receiver transferDefinition from on chain.

### DIFF
--- a/modules/router/src/forwarding.ts
+++ b/modules/router/src/forwarding.ts
@@ -147,7 +147,7 @@ export async function forwardTransferCreation(
   // Below, we figure out the correct params needed for the receiver's channel. This includes
   // potential swaps/crosschain stuff
   let recipientAmount = senderAmount;
-  let recipientTransferDefinition = senderTransferDefinition;
+  let recipientTransferType = senderTransferDefinition;
   if (recipientAssetId !== senderAssetId || recipientChainId !== senderChainId) {
     logger.warn({ method, recipientAssetId, senderAssetId, recipientChainId }, "Detected inflight swap");
     const swapRes = await getSwappedAmount(
@@ -182,7 +182,7 @@ export async function forwardTransferCreation(
         );
       }
       const senderTransferInfo = senderTransferInfoRes.getValue();
-      recipientTransferDefinition = senderTransferInfo.name;
+      recipientTransferType = senderTransferInfo.name;
     }
 
     logger.warn(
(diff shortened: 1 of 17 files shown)


modules/router/src/forwarding.ts
```diff
@@ -143,7 +147,8 @@ export async function forwardTransferCreation(
   // Below, we figure out the correct params needed for the receiver's channel. This includes
   // potential swaps/crosschain stuff
   let recipientAmount = senderAmount;
-  if (recipientAssetId !== senderAssetId) {
+  let recipientTransferDefinition = senderTransferDefinition;
+  if (recipientAssetId !== senderAssetId || recipientChainId !== senderChainId) {
```
LayneHaber, line 151: Why is there a chainId check here?

modules/router/src/forwarding.ts
```diff
@@ -156,11 +161,41 @@ export async function forwardTransferCreation(
...
+      );
+      if (senderTransferInfoRes.isError) {
+        return Result.fail(
+          new ForwardTransferError(ForwardTransferError.reasons.InvalidTransferDefinition, {
+            message: swapRes.getError()?.message,
+          }),
+        );
+      }
+      const senderTransferInfo = senderTransferInfoRes.getValue();
+      recipientTransferDefinition = senderTransferInfo.name;
```
LayneHaber, line 185: It's misleading to call this the `transferDefinition`, since that always implies it has to be an onchain address. It does work as a `type`, but this should be renamed for clarity

modules/router/src/forwarding.ts
```diff
@@ -248,7 +281,7 @@ export async function forwardTransferCreation(
         amount: recipientAmount,
         assetId: recipientAssetId,
         routingId,
-        transferDefinition,
+        transferDefinition: senderTransferDefinition,
```
LayneHaber, line 284: `transferDefinition` here is incorrect, since it's referencing the incorrect definition (though correct `type`)

LayneHaber (changes requested):

modules/router/src/forwarding.ts
rhlsthrm: Because cross-chain asset transfers still require swap profiles. Ex: does 1 mainnet ETH = 1 rinkeby ETH? If we didn't have this check, it would.

modules/router/src/forwarding.ts
rhlsthrm: Good point. Fixed.

modules/router/src/forwarding.ts
rhlsthrm: Good catch. We aren't using this ATM but will fix now.

Issue #255, opened 2020-12-09
[engine] RPC schema unit tests


LayneHaber:
RPC Schema tests in the `vector.spec.ts`

PR #261, opened 2020-12-11
new: RPC schema Unit Test

sanchaymittal:
closes: #255


modules/engine/src/testing/index.spec.ts
```diff
@@ -69,9 +81,33 @@ describe("VectorEngine", () => {
...
+        timeout: DEFAULT_CHANNEL_TIMEOUT.toString(),
+      });
+      const res = await engine.request(rpc);
+
+      console.log(res.message);
+      expect(res.message).to.contain("fail");
+    });
+    it("should fail if vector.setup fails", async () => {
+      // vectorSetup.setup.resolves(Result.fail(new Error("fail")));
+    });
```
LayneHaber, line 110: This test should either be skipped or implemented, but shouldn't appear to be passing if nothing is happening

modules/engine/src/testing/index.spec.ts
```diff
@@ -69,9 +81,33 @@ describe("VectorEngine", () => {
...
+      );
+
+      const rpc = constructRpcRequest<"chan_setup">("chan_setup", {
+        counterpartyIdentifier: counterpartyIdentifier,
+        chainId: chainId,
+        timeout: DEFAULT_CHANNEL_TIMEOUT.toString(),
+      });
+      const res = await engine.request(rpc);
+
+      console.log(res.message);
```
LayneHaber, line 105: can remove

modules/engine/src/testing/index.spec.ts
```diff
@@ -90,18 +126,715 @@ describe("VectorEngine", () => {
...
+        error: "should be equal to one of the allowed values",
+      },
+      {
+        name: "no request.method",
+        overrides: { id: 1, jsonrpc: "2.0", method: undefined, params: undefined },
+        error: "should have required property 'method'",
+      },
+      {
+        name: "malformed request.method",
+        overrides: { id: 1, jsonrpc: "2.0", method: "invalid_method", params: undefined },
```
LayneHaber, line 163: hmmm, i dont think this should succeed if the `params` can be `undefined`. i believe it should be an empty body cc @rhlsthrm

modules/engine/src/testing/index.spec.ts
```diff
@@ -90,18 +126,715 @@ describe("VectorEngine", () => {
...
+          Sinon.createStubInstance(VectorChainService),
+          chainAddresses,
+          log,
+        );
+        let rpc: EngineParams.RpcRequest = { ...test.overrides } as any;
+        try {
+          await engine.request(rpc);
+        } catch (res) {
+          expect(res.message).to.contain(test.error);
+        }
```
LayneHaber, line 188: can replace with `await expect(engine.request(rpc)).rejectedWith(test.error)`

modules/engine/src/testing/index.spec.ts
```diff
@@ -90,18 +126,715 @@ describe("VectorEngine", () => {
...
+            Sinon.createStubInstance(VectorChainService),
+            chainAddresses,
+            log,
+          );
+          let rpc: EngineParams.RpcRequest = { id: 1, jsonrpc: "2.0", ...test.overrides } as any;
+          try {
+            await engine.request(rpc);
+          } catch (res) {
+            expect(res.message).to.be.eq(test.error);
+          }
```
LayneHaber, line 835: See above replacement suggestion

LayneHaber (commented):
Leaving this as unapproved until I hear back about the empty params in the rpc schema from @rhlsthrm

modules/engine/src/testing/index.spec.ts
rhlsthrm: Some of the RPC schema is actually undefined. Maybe we should change it to be empty body only, I'm okay with that.

LayneHaber (changes requested):
Should make the `rpc.params` not allowed to be undefined (i.e. empty body instead)

modules/engine/src/testing/index.spec.ts
```diff
@@ -69,9 +81,33 @@ describe("VectorEngine", () => {
   // complex logic within the index file of the module)
 
   // NOTE: these are nice-to-haves
+
   describe.skip("setup", () => {
```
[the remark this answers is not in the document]
sanchaymittal: Still need to complete setup and withdrawal test. Will create another pr for it.

modules/engine/src/testing/index.spec.ts
sanchaymittal: Skipped, for now, removing commented code too.

modules/engine/src/testing/index.spec.ts
sanchaymittal: removed console.log

modules/engine/src/testing/index.spec.ts
sanchaymittal: Ohh noice!

modules/engine/src/testing/index.spec.ts
sanchaymittal: Done

modules/engine/src/testing/index.spec.ts
sanchaymittal: Done

LayneHaber (approved):

Issue #256, opened 2020-12-09
[engine] paramConverter tests


LayneHaber:
`paramConverter` tests skipped in `paramConverter.spec.ts`

PR #265, opened 2020-12-15
new: paramConverter test

sanchaymittal:
ref: [engine] paramConverter tests #256
closes: #256

### DIFF
--- a/modules/engine/src/paramConverter.ts
+++ b/modules/engine/src/paramConverter.ts
@@ -134,8 +134,10 @@ export async function convertWithdrawParams(
     callData,
   );
 
-  const initiatorSignature = await signer.signMessage(commitment.hashToSign());
-  if (!initiatorSignature) {
+  let initiatorSignature;
+  try {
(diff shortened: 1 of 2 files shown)


modules/engine/src/paramConverter.ts
```diff
@@ -135,6 +135,9 @@ export async function convertWithdrawParams(
   );
 
   const initiatorSignature = await signer.signMessage(commitment.hashToSign());
+  if (!initiatorSignature) {
+    return Result.fail(new Error("Signer fails to sign message"));
+  }
```
LayneHaber, line 140: We should wrap the whole thing (including the `await` in a try-catch to properly use the `Result` type

LayneHaber (changes requested):
Just the one signer change in the withdrawal converter otherwise looks great!

modules/engine/src/paramConverter.ts
sanchaymittal: done.

LayneHaber (approved):

PR #268, opened 2020-12-16
241 hardhat deploy

### LINKED ISSUE #203: [ops] Convert cli to hardhat tasks
Reduce custom code boilerplate, make use of their great devex.

Also, consider using the hardhat deploy plugin instead of our custom address book format.

### LINKED ISSUE #241: [ops] replace address book with hardhat-deploy output
More info here: https://hardhat.org/plugins/hardhat-deploy.html

hardhat-deploy plugin solves the same problem our address-book solves, using the community-maintained tool means less for us to maintain.

bohendo:
## The Problem

closes #241
closes #203


bohendo:
A few remaining TODOs:
 - Critical: make sure that deploying contracts to an arbitrary network is easy (eg via `ops/deploy-contracts.sh`). There are no tests for this & it's probably broken..
 - Performance improvement: Can (should?) we package hardhat.config.ts into a collection of js files? I wasn't able to get that working so I putt all of the contract dev deps into the ethprovider image & use the `hardhat.config.ts` file directly. This drastically increases the build time, size, and startup time of the ethprovider image. This image is only used for testing so poor performance wouldn't affect anything in prod at least. Hardhat seems to have poor support for ts->js so this might be something to do later..
 - Luxury: Should we move the hardhat deploy scripts folder back into src.ts? Feels like it belongs there but it wasn't able to find contracts until I moved it out.. this is a nice-to-have that doesn't actually affect anything.

rhlsthrm (changes requested):
Looks great so far. However I would like to see everything that's in those `actions/` turned into Hardhat [tasks](https://hardhat.org/guides/create-task.html) which encapsulate all the functionality.

Also, would like to see Typechain artifacts exported in the `contracts` package.

And can we update all `ethers` versions to the latest? I saw some did but some didn't.

bohendo:
final TODOs before this is ready to merge:
 - we need to migrate the deployment data from `address-book.json` to hardhat's file structure in `modules/contracts/deployments`
 - run a test migration to rinkeby to make sure deployment data gets updated properly

rhlsthrm (approved):

PR #292, opened 2020-12-30
make service optionally started

LayneHaber:
## The Problem

Production routers fail when not provided with a logdna key. This should not be a required service.

## The Solution

Optionally include the logdna image if the key is provided in the env


rhlsthrm (approved):
I don't understand the bash syntax but if you tested it, LGTM.

LayneHaber:
> I don't understand the bash syntax but if you tested it, LGTM.

It basically forces a test for string existence in bash. The expression will evaluate to `null` if the string isn't present. Found [here](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash)

PR #295, opened 2020-12-31
chg: crossChainTransfer function return type

sanchaymittal:


sanchaymittal:
we don't want result type return for crossChainFn

PR #296, opened 2021-01-02
token updates

LayneHaber:
## The Problem

Update token to make it workable for use as our faucet


rhlsthrm:
Going to merge this, probably something good to have.

rhlsthrm (approved):

PR #297, opened 2021-01-04
Collateral debugging

LayneHaber:

### DIFF
--- a/modules/engine/src/index.ts
+++ b/modules/engine/src/index.ts
@@ -450,7 +450,29 @@ export class VectorEngine implements IVectorEngine {
       return Result.fail(new Error(validate.errors?.map((err) => err.message).join(",")));
     }
 
-    return this.vector.deposit(params);
+    // NOTE: There is a race-condition for deposits because the onchain process
+    // is out-of-band of the protocol. For instance:
+    // 1. Alice deposits 5 onchain
+    // 2a. Alice sends single-signed update where she has reconciled the 5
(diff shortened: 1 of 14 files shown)


modules/engine/src/index.ts
```diff
@@ -450,7 +450,29 @@ export class VectorEngine implements IVectorEngine {
...
+    // 1. Alice deposits 5 onchain
+    // 2a. Alice sends single-signed update where she has reconciled the 5
+    // 2b. Bob deposits 3 onchain
+    // 3. Bob receives Alice's signature, and attempts to reconcile on their
+    //    own. Bob reconciles 8 and fails to recover Alice's signature properly
+    //    leaving all 8 out of the channel.
+
+    // There is no way to eliminate this race condition, so instead just retry
+    // depositing if a signature validation error is detected.
+    let depositRes = await this.vector.deposit(params);
```
ArjunBhuptani, line 464: I think this is an acceptable way to solve this issue -- it seems really really unlikely that Alice/Bob would continue going back and forth depositing repeatedly, so a single retry should solve the issue in most cases.

ArjunBhuptani (approved):

modules/engine/src/index.ts
LayneHaber: Yeah, helpful to have the note though otherwise I'll forget. Doubt you'll need 3 tries, but why not.

PR #299, opened 2021-01-05
Collateral issues

LayneHaber:
## The Problem

- [x]  Allow for target to be set to 0 - this makes sense because we *always* want to inflight collateralize for now
- [x]  Right now, collateralization logic errors if the `target + transfer amount > reclaim threshold`. This is incorrect. It should only error if `target > reclaim threshold`, because we should always allow a large inflight collateralization of transfer amount

### DIFF
--- a/modules/router/src/test/collateral.spec.ts
+++ b/modules/router/src/test/collateral.spec.ts
@@ -122,6 +122,37 @@ describe(testName, () => {
         amount: transferAmount.add(ethProfile.target).toString(),
       });
     });
+
+    it("should request the amount of the payment if the profile.target is 0", async () => {
+      const { channel } = createTestChannelState(UpdateType.deposit, {
+        alice: mkAddress("0xaaa"),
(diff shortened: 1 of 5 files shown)


modules/router/src/test/collateral.spec.ts
```diff
@@ -425,6 +504,32 @@ describe(testName, () => {
         expect(node.reconcileDeposit.callCount).to.be.eq(1);
       });
 
+      it("if the profile.target is 0 and a requestedAmount is supplied", async () => {
```
rhlsthrm, line 507: Would like to see this test name being more descriptive, to say `should do something`.

rhlsthrm (approved):

modules/router/src/test/collateral.spec.ts
LayneHaber: its nested in the `should work` describe

PR #309, opened 2021-01-13
308 router nats

### LINKED ISSUE #308: [router] Add NATS client
**Describe the bug**
There is no way for the router to broadcast information, like its `config`. A NATS client should be added to the router to allow for this (and other future broadcast information)

LayneHaber:
## The Problem
Fixes #308


Makefile
```diff
@@ -209,6 +209,7 @@ watch-browser-node: browser-node
 	bash ops/test-unit.sh browser-node watch
 
 test-router: router-js
+	bash ops/start-messaging.sh
```
rhlsthrm, line 212: Should we really use real messaging here? Why not the MemoryMessagingService?

modules/router/src/services/messaging.ts
```diff
@@ -0,0 +1,40 @@
...
+    const method = "publishRouterConfig";
+    const methodId = getRandomBytes32();
+    this.logger.debug({ method, methodId, config }, "Method started");
+    const subject = `${this.publicIdentifier}.config`;
+    await this.publish(subject, Result.ok(config).toJson());
+    this.logger.debug({ method, methodId }, "Method complete");
+  }
+}
+
+export const configureSubscriptions = async (
```
rhlsthrm, line 29: Why is this function necessary?

rhlsthrm (changes requested):
Let's wait on merging this until we get more info about the auctions, it seems like we don't really need this functionality once we have auctions in.

Makefile
LayneHaber: Memory messaging service is not really used anywhere, and many of the methods just throw the `Method not implemented` error.

I just copied the setup for the server-node tests which rely on a messaging service being created

modules/router/src/services/messaging.ts
LayneHaber: We will likely use it in the future to setup other nats subscriptions and broadcasts as we add more functionality to the service, similar to the `setupListeners` function, so I decided to add it in.

LayneHaber:
> Let's wait on merging this until we get more info about the auctions, it seems like we don't really need this functionality once we have auctions in.

Disagree, we've talked a few times about adding in a way to get the router config so you don't send funds to a channel if they won't support your swap

PR #310, opened 2021-01-13
Reclaim from sender channel

rhlsthrm:

### DIFF
--- a/modules/router/src/listener.ts
+++ b/modules/router/src/listener.ts
@@ -6,6 +6,7 @@ import {
   FullChannelState,
   IVectorChainReader,
   jsonifyError,
+  FullTransferState,
 } from "@connext/vector-types";
 import { getRandomBytes32 } from "@connext/vector-utils";
 import { Gauge, Registry } from "prom-client";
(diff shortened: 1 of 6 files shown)


modules/router/src/listener.ts
```diff
@@ -166,12 +167,44 @@ export async function setupListeners(
...
+      );
+
+      const transferSenderResolutionChannelAddress = res.getValue()?.channelAddress;
+      const transferSenderResolutionTransferId = res.getValue()?.transferId;
+      if (!transferSenderResolutionChannelAddress || !transferSenderResolutionTransferId) {
+        logger.warn(
+          {
+            event: EngineEvents.CONDITIONAL_TRANSFER_RESOLVED,
+            transferSenderResolutionChannelAddress,
+            transferSenderResolutionTransferId,
```
LayneHaber, line 182: should probably just log what the response is here

LayneHaber (approved):
some logging nits but otherwise lgtm

LayneHaber:
Changelog

modules/router/src/listener.ts
rhlsthrm: We do two lines above.

LayneHaber: ah should've expanded the preview, my bad

PR #313, opened 2021-01-15
Remove crosschain transfer from storage if it's canceled

rhlsthrm:


LayneHaber (approved):
Approved pending tests.

We should also log that the transfer was cancelled instead of just having it happen silently

LayneHaber:
Also, changelog

PR #314, opened 2021-01-16
Add extra gas price

rhlsthrm:
## The Problem

Add extra gwei to gas price

### DIFF
--- a/modules/contracts/src.ts/services/ethReader.ts
+++ b/modules/contracts/src.ts/services/ethReader.ts
@@ -454,6 +454,19 @@ export class EthereumChainReader implements IVectorChainReader {
     }
   }
 
+  async getGasPrice(chainId: number): Promise<Result<BigNumber, ChainError>> {
+    const provider = this.chainProviders[chainId];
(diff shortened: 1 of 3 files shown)


modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -454,6 +454,19 @@ export class EthereumChainReader implements IVectorChainReader {
...
+  async getGasPrice(chainId: number): Promise<Result<BigNumber, ChainError>> {
+    const provider = this.chainProviders[chainId];
+    if (!provider) {
+      return Result.fail(new ChainError(ChainError.reasons.ProviderNotFound));
+    }
+    try {
+      const gasPrice = await provider.getGasPrice();
+      return Result.ok(gasPrice);
+    } catch (e) {
+      return Result.fail(e);
```
LayneHaber, line 466: This isn't a chain error, could lead to `toJson` issues

LayneHaber (approved):
These should be caught by the `jsonifyError` function

PR #328, opened 2021-01-26
Idempotent init

LayneHaber:
## The Problem

- Allows `connect` on iframe to be called multiple times (i.e. with different `chainProviders`)
- Only updates disputes/sends `isAlive` messages if chain is supported
- Changes store prefix to be signer-specific, doesn't delete stores on start if signer key !== locally stored key
- Removes `crossChainTransfer` from `browser-node`


modules/server-node/examples/eth-rpc.http
```diff
@@ -18,7 +18,7 @@ Content-Type: application/json
   "method":"eth_sendTransaction",
   "params":[{
     "from": "{{sugarDaddy}}",
-    "to": "{{roger}}",
+    "to": "0x5cefD810068643D6830d10D0f3b25d82fC0E8763",
```
rhlsthrm, line 21: Revert this.

modules/test-ui/src/App.tsx
```diff
@@ -37,40 +37,40 @@ function App() {
...
+      const _channels = (
+        await Promise.all(
+          channelAddresses.map(async (c) => {
+            const channelRes = await client.getStateChannel({ channelAddress: c });
+            console.log("Channel found in store:", channelRes.getValue());
+            const channelVal = channelRes.getValue() as FullChannelState;
+            return channelVal;
+          }),
+        )
+      ).filter((chan) => supportedChains.includes(chan.networkContext.chainId));
```
rhlsthrm, line 73: Not sure if we should do this in test-ui, what if we don't have supported chains (everything should work without it).

rhlsthrm (approved):
Pending tests and minor feedback.

modules/test-ui/src/App.tsx
LayneHaber: This was causing errors for me when i only wanted to connect to one provider.

Transfers won't work without it, since you can't call chain service functions (i.e. `create` or `resolve`)

PR #329, opened 2021-01-26
new: add signature as param for iframe

sanchaymittal:


rhlsthrm:
Already done in a separate branch.

Issue #332, opened 2021-01-27
[router] Observability/telemetry


ArjunBhuptani:
We disabled telemetry because it was overloading routers with db requests. However, we'll need to be able to track metrics around costs and value locked in channels for the first few routers to be able to understand what's going on with them at scale.

sanchaymittal:
Start with
- signer address balances
- engine events

PR #337, opened 2021-01-28
332 add metrics collector

rhlsthrm:
## The Problem

First part of #332 observability

## The Solution

TODO:
- [x] Metrics collector ops and hook up to Prometheus.
- [x] Query signer balance in metrics collector.
- [x] Show signer balances on dashboard.
- [ ] Add metrics registration for router events, remove all other metrics. Note: this requires bubbling the rest of the events up to the router, right now it's only listening to a subset. I think it makes sense to do this, then the node doesn't have to worry about this.
- [ ] Show event metrics on dashboard.


rhlsthrm:
Separate router events into new PR.

PR #338, opened 2021-01-29
add: use gasNow for ethereum mainnet gas price.

### LINKED ISSUE #333: Switch to using gasNow instead of Ethers for gas price estimation

sanchaymittal:
closes #333

### DIFF
--- a/modules/contracts/src.ts/services/ethReader.ts
+++ b/modules/contracts/src.ts/services/ethReader.ts
@@ -454,11 +454,17 @@ export class EthereumChainReader implements IVectorChainReader {
       return Result.fail(new ChainError(ChainError.reasons.ProviderNotFound));
     }
     try {
-      let gasPrice = await provider.getGasPrice();
+      let gasPrice;
       if (chainId === 1) {
-        const gasNowResponse = await axios.get(`https://www.gasnow.org/api/v3/gas/price`);
-        const gasNowData = gasNowResponse.data;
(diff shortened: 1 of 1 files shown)


modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -453,7 +454,12 @@ export class EthereumChainReader implements IVectorChainReader {
       return Result.fail(new ChainError(ChainError.reasons.ProviderNotFound));
     }
     try {
-      const gasPrice = await provider.getGasPrice();
+      let gasPrice = await provider.getGasPrice();
```
rhlsthrm, line 457: Put this in the else so that you don't need to call the provider function if you're using gasnow.

modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -453,7 +454,12 @@ export class EthereumChainReader implements IVectorChainReader {
...
+      if (chainId === 1) {
+        const gasNowResponse = await axios.get(`https://www.gasnow.org/api/v3/gas/price`);
```
rhlsthrm, line 459: Wrap this in a try-catch and fall back to `provider.getGasPrice();`.

rhlsthrm (changes requested):
Nice work, but needs a couple changes.

LayneHaber (approved):

PR #370, opened 2021-02-10
368 resolved transfers channel

### LINKED ISSUE #359: [store] Withdrawal tx hashes
**Describe the bug**
Withdrawal transfers are not associated with a transaction anywhere in the store. This makes determining which withdrawals have been successfully submitted and which haven't a nightmare.

### LINKED ISSUE #368: [server-node] Resolved transfers have no channel relation
**Describe the bug**
Resolved transfers remove their `channel` relation and keep a `channelAddressId`

This is done to keep track of the `activeTransfers` but makes it difficult to get properties needed from the channel for the withdrawal commitment, see [here](https://github.com/connext/vector/blob/16c9de9ec675cc9a227afcffe44e29c25adbb49c/modules/server-node/src/services/store.ts#L392-L393). 

This causes problems when you attempt to retry withdrawal submissions, and will block #354 and #358 

Instead, the channel should keep a reference to active transfers

LayneHaber:
## The Problem

Fixes #368
Fixes #359


LayneHaber:
it is a bit confusing how the onchain transaction has the `transactionHash` as the id field but then the unique assertion is on the `[transactionHash, chainId]` combo

im not sure if thats a problem or not with how the tx is associated

rhlsthrm (approved):

rhlsthrm:
Good pending build fixes and tests.

PR #382, opened 2021-02-15
Don't log secrets to stdout

3nprob:
## The Problem
All secrets in config (adminToken, wallet mnemonic in cleartext) as well as database credentials, are logged to stdout as part of normal operations. This makes logs extremely sensitive - just the startup logs would be enough to drain the wallet, even without any internal network access.

## The Solution
Remove the logging for database url and full config.

It can still be desirable to log the rest of the config on startup, which can be done in a future PR by spoecifically redacting the sensitive values.


rhlsthrm:
Thanks for this! Before we merge this I'd like to as you mention add back the config logging with the sensitive info redacted.

rhlsthrm (approved):

PR #391, opened 2021-02-18
V0 Router fees implementation

### LINKED ISSUE #360: [router, config] Scope and implement fees
We need to implement at least some basic fee capture mechanisms to pass costs of operating a router down to the end user. 

Right now we have: 
(1) Spreads on swap rates
(2) Fees on withdraw

We probably also need:
- A way to charge fees during transfers (e.g. xud)
- A way to estimate fees and show that to the user in the UI

We can subsidize the cost of going from Ethereum -> L2, but will likely need to show at least *some* fees for L2 -> Ethereum given where gas is at.

rhlsthrm:
## The Problem

Fixes #360 

## The Solution

TODO:
- [x] Add config items for flatFee, percentageFee, and dynamicFee on both allowedSwaps and base (if not a swap).
- [x] Calculate dynamicFee based on estimated gas usage + oracle pricing.
- [x] Add `getFeeQuote` endpoint to router.
- [x] Unit tests.
- [ ] Docs.

Note: We might want to cache the result of the dynamic fee calculation so that we can avoid calling it every time. Look into https://github.com/node-cache/node-cache.


LayneHaber (approved):
Dont release as non-beta until more tests are added

PR #394, opened 2021-02-22
Router restart fixes

LayneHaber:
## The Problem

No way to handle transfers that were dropped due to router restart


rhlsthrm (approved):
Looks good to push out and test with the users that are seeing issues.

PR #398, opened 2021-02-24
[DNM] Add logOperationTime util with implementation in protocol module

jakekidd:
## The Problem

- We need accurate logging to document time it's taking for methods to execute, starting with protocol module

## The Solution

- Implemented logOperationTime, a util method
- Used at start and completion of a method to log start and time it took to execute, respectively

- Used performance(.mark()) as opposed to Date.now() for accuracy


package.json
```diff
@@ -23,6 +23,5 @@
     "eslint-plugin-import": "2.22.1",
     "eslint-plugin-prettier": "3.1.4",
     "pino-pretty": "4.3.0"
-  },
-  "dependencies": {}
+  }
```
jakekidd, line 26: Not sure why this is getting removed for me - might be result of using later version of node. Can fix before I check in.

modules/utils/src/logOperationTime.ts
```diff
@@ -0,0 +1,33 @@
...
+/// perf_hooks (performance) lib provided by node to calculate this - much more accurate than
+/// Date.now(). On first call, a unique label string is generated and returned. On second call,
+/// same label string should be passed in to measure and log the diff in time (i.e. amount of
+/// time method took to execute in ms).
+export const logOperationTime = (log: pino.BaseLogger, method: string, label?: string): string => {
+  if (label) {
+    const stopLabel = `${label}-stop`;
+    const timestamp = performance.mark(stopLabel);
+    const diff = performance.measure(method, label, stopLabel);
+    const message = `${method} : finished @ ${timestamp}; took ${diff} ms`;
```
LayneHaber, line 15: Probably only need the diff here since pino adds timestamps

modules/utils/src/logOperationTime.ts
```diff
@@ -0,0 +1,33 @@
...
+import { getRandomBytes32 } from "@connext/vector-utils";
+
+/// This method is used to log the start and end time of an executed method. We are using the
+/// perf_hooks (performance) lib provided by node to calculate this - much more accurate than
+/// Date.now(). On first call, a unique label string is generated and returned. On second call,
+/// same label string should be passed in to measure and log the diff in time (i.e. amount of
+/// time method took to execute in ms).
+export const logOperationTime = (log: pino.BaseLogger, method: string, label?: string): string => {
+  if (label) {
+    const stopLabel = `${label}-stop`;
```
LayneHaber, line 12: What would happen if there are label collisions?

modules/protocol/src/vector.ts
```diff
@@ -199,7 +201,7 @@ export class Vector implements IVectorProtocol {
         }
         const method = "onReceiveProtocolMessage";
         const methodId = getRandomBytes32();
-        this.logger.debug({ method, methodId }, "Method start");
+        const label = logOperationTime(this.logger, method);
```
LayneHaber, line 204: We will likely want to keep the `methodId` to log throughout the other parts of the method so the logs are easily searchable. Maybe we should use the `label` as the `methodId` instead.

General purpose of the `methodId` is to get a unique strings so that you can easily identify all logs that came from a given method when searching in log dna or whatever other log parser

jakekidd: Sure. This functionality is already internal to the `logOperationTime` method. I actually had it like that at first, but worried that not all methods would follow this standard of generating a methodId. Will add this back tho, thanks

modules/utils/src/logOperationTime.ts
jakekidd: Yep, this is from earlier iteration, good catch

modules/utils/src/logOperationTime.ts
jakekidd: Currently using a unique identifier below as part of generated label to prevent that, but this will be replaced by passed in methodId

package.json
rhlsthrm: Not really an issue, we can push this change. We will be updating to Node 14 soon anyways.

rhlsthrm:
We should make this time logging part of the "debug" log config. IMO it should be in the code and be able to be enabled with a "switch flip" of sorts.

jakekidd:
> We should make this time logging part of the "debug" log config. IMO it should be in the code and be able to be enabled with a "switch flip" of sorts.

Can you clarify what you mean by "debug" log config? Is there a configuration somewhere I should be aware of? The way it is currently, unless the method takes an inordinate amount of time to execute, it will always log it on debug level. Takes too long and it logs it as info, then warning

rhlsthrm:
> Can you clarify what you mean by "debug" log config? Is there a configuration somewhere I should be aware of? The way it is currently, unless the method takes an inordinate amount of time to execute, it will always log it on debug level. Takes too long and it logs it as info, then warning

https://docs.connext.network/configuring-a-router here's the config.

Your implementation sounds good. The main thing I was trying to get at is we should have the ability to turn this on/off in the config. So as part of this PR I think we should add a config item called `enableTimestampLogging` or something which can turn the whole thing on/off.

jakekidd:
> https://docs.connext.network/configuring-a-router

Ah, I misread your first comment. I thought you said it should NOT be able to be enabled by switch flip of sorts , haha.

Yes, agreed. I will target that functionality for this PR, thanks. (Meaning I will include addition of this config item)

PR #413, opened 2021-03-05
Rebalance l2 --> l1

### LINKED ISSUE #412: [router] Rebalance withdrawals
**Describe the bug**
We can rebalance from l1 --> l2 but not l2 --> l1, fix

Flow:
- [ ] add schema/db table for active rebalances with a status
- [ ] on each rebalance interval, check rebalance thresholds for allowedSwaps
- [ ] if out of threshold, check if active rebalance for that swap exists
- [ ] if it does not, create one and go through the approve/execute flow
- [ ] if it does exist, check status. if status == {completed: true} and if there is a tx in the response (is the case with matic withdraw), send the tx, and mark as completed. if no tx in the response, mark as completed immediately
- [ ] if it is not completed, wait for next poll

LayneHaber:
## The Problem

Closes #412


LayneHaber:
Next steps:
(1) update chain rebalancer service so it works for `AddressZero` (pretty sure i did this already, pushed to main)
(2) move types from fustily repo into vector package
(3) write the store migrations/update the router store schema
(4) fix the build in the router (likely some errors in tests and errors themselves)
(5) test the rebalancer service on the router

modules/router/src/services/store.ts
```diff
@@ -111,4 +134,88 @@ export class PrismaStore implements IRouterStore {
...
+    const now = new Date(timeElapsed);
+    await this.prisma.autoRebalance.update({
+      where: {
+        id: record.id
+      },
+      data: {
+        // Update the 'updatedAt' time.
+        updatedAt: now,
+        // Core data updates:
+        // TODO: Should all of these items be updatable?
```
rhlsthrm, line 207: I don't think the config items should be updated, just the status and hashes.

modules/router/prisma-postgres/schema.prisma
```diff
@@ -17,3 +17,33 @@ model QueuedUpdate {
...
+  flatFee                 String?
+  gasSubsidyPercentage    Int?
+
+
+  createdAt               DateTime @default(now())
+  updatedAt               DateTime @default(now())
+
+  @@unique([fromChainId, toChainId, fromAssetId, toAssetId])
+  @@unique(approveHash)
+  @@unique(completeHash)
```
rhlsthrm, line 47: I think we'll need to remove these unique constraints. Below you mention that the completion tx hash won't be present for the L1->L2 rebalance. If that's the case it will violate the constraint. I think it's acceptable to loosen the restriction and remove the uniques on the hashes. I appreciate the specificity though 👍.

rhlsthrm (changes requested):
This looks good so far but looks incomplete. We need to address the TODOs about saving to the DB. Also, see my comment about the unique constraint.

modules/router/prisma-postgres/schema.prisma
jakekidd: Should we add `@@unique()` for ID column above? Wondering why that's not a thing.

rhlsthrm: `id` columns are unique by default since they are primary key.

modules/router/src/services/store.ts
```diff
@@ -2,6 +2,7 @@ import { NodeParams, AllowedSwap } from "@connext/vector-types";
 
 import { PrismaClient } from "../generated/db-client";
 import { getConfig } from "../config";
+import { exception } from "node:console";
```
rhlsthrm, line 5: What's this?

rhlsthrm (approved):

rhlsthrm:
Pending trio tests.

PR #457, opened 2021-03-09
454 expose getters

### LINKED ISSUE #454: [server-node, browser-node, engine] Expose withdrawal commitment getters in server-node and browser-node interfaces

LayneHaber:
## The Problem

Fixes #454


LayneHaber:
NOTE: as submitted will returned `undefined` if the commitment is not found. As I was doing this PR I noticed we were a bit inconsistent with how we were handling `undefined` types on the server-node itself.

rhlsthrm (approved):

rhlsthrm:
Merged without docs changes, hoping @LayneHaber will do this after the fact! No longer will be allowed though.

Issue #378, opened 2021-02-11
[types] Make timeouts actually optional


ArjunBhuptani:
Timeouts on `setup` and `create` should be optional but aren't. We shouldn't be having users define their own timeouts unless they know exactly what they're doing.

rhlsthrm:
To add to this issue, I found that withdraw param converter isn't actually passing the timeout through so even if it gets set it gets set to the default transfer timeout.

jakekidd:
While I haven't handled second issue mentioned by @rhlsthrm above, here is PR: https://github.com/connext/vector/pull/463

PR #463, opened 2021-03-15
modify : SetupParams and CreateParams for transfer to take optional timeout

jakekidd:
## The Problem
From [#378](https://github.com/connext/vector/issues/378):
>Timeouts on setup and create should be optional but aren't. We shouldn't be having users define their own timeouts unless they know exactly what they're doing.

Additional problem:
>To add to this issue, I found that withdraw param converter isn't actually passing the timeout through so even if it gets set it gets set to the default transfer timeout.

## The Solution
[types] Made timeout param optional in Setup and Create types (as well as schemas).

Where necessary, handled optional timeouts as passed from source, and defaulted to `DEFAULT_TRANSFER_TIMEOUT`.

Archbee changes:
Changes made to [API Reference](https://app.archbee.io/docs/R6w14eIlq8wdqU9likOX_/nDN6LfXhF7Xh2UIK1Ui9G):
  - Added `timeout: Type.Optional(TIntegerString)` under `WithdrawParamsSchema` under *withdraw* section
  - Set `timeout` param to be optional (`timeout: Type.Optional(TIntegerString)`) under *setup* section


modules/types/src/channel.ts
```diff
@@ -35,7 +35,7 @@ export type CreateTransferParams = {
   assetId: string;
   transferDefinition: string;
   transferInitialState: TransferState;
-  timeout: string;
+  timeout?: string;
```
LayneHaber, line 38: Here and in the `SetupParams` we should enforce the presence. It's in the engine `schema` that this should happen, and these should be autofilled so that in the protocol it is required

modules/types/src/schemas/protocol.ts
```diff
@@ -20,7 +20,7 @@ import {
 
 // Setup
 const SetupProtocolParamsSchema = Type.Object({
-  timeout: TIntegerString,
+  timeout: Type.Optional(TIntegerString),
```
LayneHaber, line 23: Should be in the engine instead of in the protocol

modules/types/src/schemas/protocol.ts
```diff
@@ -40,7 +40,7 @@ const CreateProtocolParamsSchema = Type.Object({
   assetId: TAddress,
   transferDefinition: TAddress,
   transferInitialState: TransferStateSchema,
-  timeout: TIntegerString,
+  timeout: Type.Optional(TIntegerString),
```
LayneHaber, line 43: Should be in the engine instead of the protocol

modules/utils/src/transfers.ts
```diff
@@ -19,11 +20,11 @@ export const getTransferId = (
   channelAddress: Address,
   channelNonce: string,
   transferDefinition: Address,
-  transferTimeout: string,
+  transferTimeout?: string,
```
LayneHaber, line 23: should be required here, optional in the API params for the engine only

LayneHaber (changes requested):
Couple of things:

1. This makes the `protocol` parameters optional, when really it should be handled in the `paramConverter` portion of the engine (and the protocol itself should enforce the presence and validity of these timeouts through its schema)
2. You should use the `??` syntax instead of the ternary operator to avoid any weird problems caused by null-ish values
3. The timeout should already be optional for creating transfers, its really only for `setup` that you need to add the defaults and the default used should be the `DEFAULT_CHANNEL_TIMEOUT` constant
4. Please update the `changelog.md` (we either do this here or on merge, but realistically we forget a lot on merge so probably better to just start being strict about this lol)
5. Please link to the updated staging archbee docs in the PR body :)

In general, the `protocol` module is pretty settled so if you're touching something there it should be dispute related, or you may be in the wrong place. We should have been more clear about that in the issue description, but at least you got a tour!

jakekidd:
> Couple of things:
> 
> 1. This makes the `protocol` parameters optional, when really it should be handled in the `paramConverter` portion of the engine (and the protocol itself should enforce the presence and validity of these timeouts through its schema)
> 2. You should use the `??` syntax instead of the ternary operator to avoid any weird problems caused by null-ish values
> 3. The timeout should already be optional for creating transfers, its really only for `setup` that you need to add the defaults and the default used should be the `DEFAULT_CHANNEL_TIMEOUT` constant
> 4. Please update the `changelog.md` (we either do this here or on merge, but realistically we forget a lot on merge so probably better to just start being strict about this lol)
> 5. Please link to the updated staging archbee docs in the PR body :)
> 
> In general, the `protocol` module is pretty settled so if you're touching something there it should be dispute related, or you may be in the wrong place. We should have been more clear about that in the issue description, but at least you got a tour!

Thanks for the detailed explanation! On this now.

LayneHaber (approved):
Nicely done!

PR #466, opened 2021-03-16
Fix error message

rhlsthrm:
## The Problem

Resolves https://github.com/connext/vector-modal/issues/69.


LayneHaber (approved):
So sad, good catch

LayneHaber:
wow very sad

PR #474, opened 2021-03-18
467 normalize fees

### LINKED ISSUE #467: [router] Normalize observed fees/costs
**Describe the bug**
To generate a good idea of fees vs. costs for router operators, all fees and gas must be normalized to the same unit (i.e. eth denominated)

LayneHaber:
## The Problem

Fixes #467 

Important to note -- we are using the json file to calculate mainnet equivalent fees


rhlsthrm (changes requested):
I think my suggestion to use chains.json as a manual mapping of L2 assets to mainnet assets will work. I will add some of this to the PR.

rhlsthrm (approved):
This will require modal changes also.

Issue #246, opened 2020-12-08
[server-node, browser-node] Combine store tests


rhlsthrm:
Instead of copy-pasting store tests we need to combine them in one place.

jakekidd:
https://github.com/connext/vector/pull/481

jakekidd:
v__

PR #481, opened 2021-03-19
add : testStore, call in browser-node store.spec.ts

jakekidd:
## The Problem
https://github.com/connext/vector/issues/246
>Instead of copy-pasting store tests we need to combine them in one place.

## The Solution
Made a testStore method in utils that takes either BrowserStore or PrismaStore (as well as callback to handle init) in utils. The `spec.ts` for both the browser-node and server-node test will share this util.


rhlsthrm (approved):
Nice work!

PR #491, opened 2021-03-22
add : capability to test server-node store with testStore utility

jakekidd:
## The Problem
Issue has been deleted, but this is a follow-up to get the second half of tests working for browser-node/server-node.

## The Solution
 With this update, server-node will also use the `testStore` utility in vector-utils.


modules/server-node/src/services/store.spec.ts
```diff
@@ -1,767 +1,682 @@
...
-  });
-});
+const name = "PrismaStore";
+const { log } = getTestLoggers(name);
+const pubId = mkPublicIdentifier();
+
+testStore(name, () => new PrismaStore(config.dbUrl))
+
+
+// import {
```
jakekidd, line 17: Oops. Will remove this commented code

LayneHaber (approved):

PR #495, opened 2021-03-22
add : retryWrapper for all necessary methods in EthereumChainReader

### LINKED ISSUE #465: [contracts] Retry all calls in chainReader
**Describe the bug**
Dealing with eth nodes is often unreliable. The `chainReader` should retry these calls, especially since no gas is wasted. Ideally we'd sort by error messages, but we don't even need the implementation to be that complex.

jakekidd:
## The Problem
Fixes #465 
>Dealing with eth nodes is often unreliable. The chainReader should retry these calls, especially since no gas is wasted. Ideally we'd sort by error messages, but we don't even need the implementation to be that complex.

## The Solution
Add `retryWrapper` method and implement across the board. Have a constant defined amount # of retries.

### DIFF
--- a/modules/contracts/src.ts/services/ethReader.ts
+++ b/modules/contracts/src.ts/services/ethReader.ts
@@ -293,101 +293,98 @@ export class EthereumChainReader implements IVectorChainReader {
     chainId: number,
     bytecode?: string,
   ): Promise<Result<boolean, ChainError>> {
-    const provider = this.chainProviders[chainId];
-    if (!provider) {
-      return Result.fail(new ChainError(ChainError.reasons.ProviderNotFound));
-    }
-    // Get encoding
-    const registryRes = await this.getRegisteredTransferByDefinition(
-      transferDefinition,
-      transferRegistryAddress,
-      chainId,
-      bytecode,
-    );
-    if (registryRes.isError) {
-      return Result.fail(registryRes.getError()!);
-    }
-    // Try to encode
-    let encodedState: string;
-    let encodedBalance: string;
-    try {
-      encodedState = encodeTransferState(initialState, registryRes.getValue().stateEncoding);
-      encodedBalance = encodeBalance(balance);
-    } catch (e) {
-      return Result.fail(e);
-    }
-    const contract = new Contract(transferDefinition, TransferDefinition.abi, provider);
-    if (bytecode) {
-      const evmRes = this.tryEvm(
-        contract.interface.encodeFunctionData("create", [encodedBalance, encodedState]),
+    return await this.retryWrapper<boolean>(chainId,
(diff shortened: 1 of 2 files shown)


modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -574,40 +558,61 @@ export class EthereumChainReader implements IVectorChainReader {
...
+        return Result.ok(cleaned);
+      }
+    );
+  }
+
+  private async retryWrapper<T>(
+    chainId: number,
+    targetMethod: (provider: JsonRpcProvider) => Promise<Result<T, ChainError>>
+  ): Promise<Result<T, ChainError>> {
+    // TODO : Should we retry this attempt to get provider as well?
```
LayneHaber, line 602: No we can leave this outside of the retries

modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -574,40 +558,61 @@ export class EthereumChainReader implements IVectorChainReader {
...
-      }
-    }
-    if (!registered) {
-      try {
-        registered = await registry.getTransferDefinitions();
-      } catch (e) {
-        return Result.fail(new ChainError(e.message, { chainId, transferRegistry }));
-      }
+    // TODO: Move this constant where it belongs.
+    const MAX_RETRIES = 5;
```
LayneHaber, line 608: Let's just move this to `modules/types/constants.ts` and name it something more distinct

modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -574,40 +558,61 @@ export class EthereumChainReader implements IVectorChainReader {
...
+    let res = await targetMethod(provider);
+    let retries = 0;
+
+    while (res.getError() && retries < MAX_RETRIES) {
```
LayneHaber, line 612: can use the `isError` flag here

LayneHaber (commented):
Looks good, I would take care of the `TODO` for the constant and if possible add at least one test where the provider throws on the first call and resolves on the second.

modules/contracts/src.ts/services/ethReader.ts
jakekidd: done, named `ETH_READER_MAX_RETRIES`. If that's good, pls resolve comment, thx
\

modules/contracts/src.ts/services/ethReader.ts
jakekidd: Oooh good to know

modules/contracts/src.ts/services/ethReader.ts
jakekidd: Copy

LayneHaber (approved):

PR #505, opened 2021-03-25
Disputes

### LINKED ISSUE #518: [protocol, contracts, server-node] MVP disputes
**Describe the bug**
MVP of offchain dispute implementations

rhlsthrm:
## The Problem

Closes #518


LayneHaber:
[Here](https://www.notion.so/connext/Disputes-3fcf7f8b74344fe7acc2e69c91a5edee) is an update on what is and isn't included in the MVP implementation/this PR

modules/engine/src/listeners.ts
```diff
@@ -783,7 +801,7 @@ async function handleConditionalTransferResolution(
   };
   evts[EngineEvents.CONDITIONAL_TRANSFER_RESOLVED].post(payload);
 }
-
+``;
```
rhlsthrm, line 804: ?

rhlsthrm (commented):
Looks good. However, this is a huge PR. I think the best way to do this is to do a live code review @LayneHaber.

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -97,6 +101,44 @@ export class EthereumChainService extends EthereumChainReader implements IVector
     if (!channelState.latestUpdate.aliceSignature || !channelState.latestUpdate.bobSignature) {
       return Result.fail(new ChainError(ChainError.reasons.MissingSigs));
     }
+
+    const code = await this.getCode(channelState.channelAddress, channelState.networkContext.chainId);
+    if (code.isError) {
+      return Result.fail(code.getError()!);
+    }
+    if (code.getValue() === "0x") {
```
jakekidd, line 109: what is it about this condition that prompts us to deploy the channel? why do we deploy a channel just to defund it due to a dispute?

rhlsthrm: You need the channel deployed to call any of the methods on the contracts. If it's not deployed it's just an address.

LayneHaber (approved):

PR #509, opened 2021-03-26
Implement chain provider class with fallback capabilities

jakekidd:
## The Problem
Issue #410: https://github.com/connext/vector/issues/410

## The Solution
The user should be able to specify in their router's config multiple provider addresses which can be used to fallback to.

For fallback functionality, we are leaning on ethersproject FallbackProvider, but have made a custom extending class `ChainProvider` which wraps that functionality but also enforces we are using JsonRpcProvider.

Here we've implemented this change with backwards compatability by using comma-separated values. Example for `chainProviders` in config:
```
{
  "1337": "http://localhost:8545,http://localhost:8546",
  "1338": "http://localhost:8547"
}
```

TODO: Link change in docs where we've updated the config API.

### DIFF
--- a/modules/types/src/network.ts
+++ b/modules/types/src/network.ts
@@ -1,9 +1,53 @@
-import { JsonRpcProvider } from "@ethersproject/providers";
+import { JsonRpcProvider, FallbackProvider } from "@ethersproject/providers";
 
 export type ChainProviders = {
-  [chainId: number]: string;
+  [chainId: number]: string[]
 };
 
 export type HydratedProviders = {
-  [chainId: number]: JsonRpcProvider;
+  [chainId: number]: ChainProvider;
 };
+
+/* Represents an aggregate of providers for a particular chain. Leverages functionality from
+*  @ethersproject/providers/FallbackProvider in order to fallback to other providers in the
+*  event of failed requests.
+*/
+export class ChainProvider extends FallbackProvider {
+    readonly chainId: number;
+    readonly providerUrls: string[];
+
+    constructor(chainId: number, providers: string[] | JsonRpcProvider[], stallTimeout?: number) {
+      // We'll collect all the provider URLs as we hydrate each provider.
+      var providerUrls: string[] = [];
+      super(
+        // Map the provider URLs into JsonRpcProviders 
+        providers.map((provider: string | JsonRpcProvider, priority: number) => {
+          const hydratedProvider = (typeof(provider) === "string") ? new JsonRpcProvider(provider, chainId) : provider;
+          providerUrls.push(hydratedProvider.connection.url);
+          return {
+            provider: hydratedProvider,
+            // Invert priority as higher values are used first.
+            priority: -priority,
+            // Timeout before also triggering the next provider; this does not stop
+            // this provider and if its result comes back before a quorum is reached
+            // it will be incorporated into the vote
+            // - lower values will cause more network traffic but may result in a
+            //   faster retult.
+            // TODO: Should we have our own default timeout defined, as well as a config option for this?
+            // Default timeout is written as either 2sec or .75sec (in @ethers-project/fallback-provider.ts):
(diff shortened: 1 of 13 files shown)


rhlsthrm (approved):
This looks great, nice work.

rhlsthrm:
Pending tests.

modules/contracts/src.sol/CMCCore.sol
```diff
@@ -66,7 +66,7 @@ contract CMCCore is ReentrancyGuard, ICMCCore {
     }
 
     /// @notice A getter function for the bob of the multisig
-    /// @return Alice's signer address
+    /// @return Bob's signer address
```
jakekidd, line 69: @LayneHaber @rhlsthrm idk when this happened, probably when I was snooping around codebase and noticed this weirdness. Lmk if you want me to remove as irrelevant to PR... but also it's just a comment haha

jakekidd: nvm removed

rhlsthrm (commented):
Not fully clear on why we need a custom class yet.

modules/types/src/network.ts
```diff
@@ -1,9 +1,306 @@
...
+    //   console.log("DEBUG MESSAGE:")
+    //   console.log(key, this._provider[key]);
+    //   if (typeof this._provider[key] == "function") {
+    //     Object.assign(this, {
+    //       key: this._provider[key]
+    //     });
+    //     console.log(this[key]);
+    //   }
+    // }
+  }
```
jakekidd, line 112: Not needed I think; remove

modules/utils/src/eth.ts
```diff
@@ -20,10 +26,21 @@ export const getGasPrice = async (provider: Provider, providedChainId?: number):
...
+    }
+  );
+  return chainProviders
+}
+
+export const hydrateProviders = (chainProviders: ChainRpcProviders): HydratedProviders => {
+  const hydratedProviders: { [url: string]: ChainRpcProvider } = {};
   Object.entries(chainProviders).map(([chainId, url]) => {
-    hydratedProviders[chainId] = new StaticJsonRpcProvider(url as string, parseInt(chainId));
+    hydratedProviders[chainId] = new ChainRpcProvider(parseInt(chainId), url);
```
jakekidd, line 43: Noting that we replaced StaticJsonRpcProvider here - is there something special / different about StaticJsonRpcProvider vs JsonRpcProvider? Do we need to implement additional functionality or a 'Static' version of ChainRpcProvider?

modules/utils/src/eth.ts
```diff
@@ -1,16 +1,22 @@
-import { ChainProviders, HydratedProviders } from "@connext/vector-types";
+import { ChainRpcProviders, HydratedProviders } from "@connext/vector-types";
 import { Provider } from "@ethersproject/abstract-provider";
 import { BigNumber } from "@ethersproject/bignumber";
-import { JsonRpcProvider, StaticJsonRpcProvider } from "@ethersproject/providers";
+import { JsonRpcProvider } from "@ethersproject/providers";
+import { ChainRpcProvider } from "@connext/vector-types";
```
jakekidd, line 5: remove duplicate import

modules/utils/src/eth.ts
rhlsthrm: Hm, we switched to `StaticJsonRpcProvider` because it cuts down on the number of RPC calls. We should just instantiate as `StaticJsonRpcProvider`s under the hood. Is that possible?

jakekidd: Yes but lower priority atm

jakekidd: Do you mean replace JsonRpcProvider with StaticJsonRpcProvider entirely? Or just in this one use case?

rhlsthrm: Yes replace entirely. That's what we already did AFAIK.

PR #528, opened 2021-03-31
Fix autorebalance bug

### LINKED ISSUE #519: [router] Autorebalance fixed
**Describe the bug**
Autorebalancing fixed between mainnet and matic

rhlsthrm:
## The Problem

Fixes #519 

## The Solution

Fix store methods.

TODO:
- [ ] Save tx hashes. Must be saved right after submission not confirmation.
- [ ] Check tx hash on each step to make sure it's confirmed (i.e. what if we drop out during approve and have the hash but it's not confirmed?).


modules/router/src/services/autoRebalance.ts
```diff
@@ -183,14 +184,15 @@ export const rebalanceIfNeeded = async (
...
+  console.log("latest: ", latest);
   if (!latest) {
     // set dummy value
     latest = {
       status: RouterRebalanceStatus.COMPLETE,
       swap,
       approveHash: undefined,
       executeHash: undefined,
-      id: getRandomBytes32(), // dummy value, set by db
+      id: uuidv4(), // dummy value, set by db
```
LayneHaber, line 195: should there be a `completeHash` here as well?

LayneHaber (approved):

modules/router/src/services/store.ts
```diff
@@ -90,6 +90,7 @@ export class PrismaStore implements IRouterStore {
 
   async clear(): Promise<void> {
     await this.prisma.queuedUpdate.deleteMany({});
+    await this.prisma.autoRebalance.deleteMany({});
```
jakekidd, line 93: Oops. Great catch

jakekidd (approved):

PR #532, opened 2021-03-31
Fix double collateral issue

rhlsthrm:
## The Problem

Found an issue where double collateralization can occur.

## The Solution

Fix issue.


modules/router/src/services/collateral.ts
```diff
@@ -338,22 +339,28 @@ export const requestCollateral = async (
   const processed =
     participant === "alice" ? channel.processedDepositsA[assetIdx] : channel.processedDepositsB[assetIdx];
   const amountToDeposit = BigNumber.from(target).sub(myBalance);
-  const reconcilable = totalDeposited.getValue().sub(processed ?? "0");
+  const reconcilable = totalDeposited
+    .getValue()
+    .sub(processed ?? "0")
+    .lt(Zero)
+    ? Zero
+    : totalDeposited.getValue().sub(processed ?? "0");
```
LayneHaber, line 347: This will never be less than 0 unless something has gone horribly wrong

modules/router/src/test/collateral.spec.ts
```diff
@@ -771,5 +771,48 @@ describe(testName, () => {
         expect(node.reconcileDeposit.callCount).to.be.eq(2);
       });
     });
+
+    it("should collateralize difference between owed and target", async () => {
+      const { channel } = createTestChannelState(UpdateType.deposit, {
+        aliceIdentifier: routerPublicIdentifier,
+        processedDepositsA: ["0"],
+        assetIds: [AddressZero],
+        balances: [{ amount: ["0", "0"], to: [AddressZero, AddressZero] }],
```
LayneHaber, line 780: the `to` in the balance here should be the channel participants, not the `assetId`

LayneHaber (changes requested):

PR #539, opened 2021-04-02
Dont register dispute

rhlsthrm:
## The Solution

TODO:

- [ ] Expose API for syncing disputes.

### DIFF
--- a/modules/engine/src/index.ts
+++ b/modules/engine/src/index.ts
@@ -1528,6 +1528,11 @@ export class VectorEngine implements IVectorEngine {
     return Result.ok(results);
   }
 
+  private async syncDisputes(): Promise<Result<void, EngineError>> {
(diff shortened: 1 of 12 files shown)


modules/engine/src/index.ts
```diff
@@ -1528,6 +1528,11 @@ export class VectorEngine implements IVectorEngine {
     return Result.ok(results);
   }
 
+  private async syncDisputes(): Promise<Result<void, EngineError>> {
+    await this.vector.syncDisputes();
+    return Result.ok(undefined);
+  }
```
LayneHaber, line 1534: This should be wrapped in a  `try-catch` block or you'll get `UnhandledPromiseRejection` errors

LayneHaber (changes requested):
If router is gating this behind `skipCheckIn` in the config, it should call this on startup

PR #540, opened 2021-04-02
Handle withdrawal cancellations

rhlsthrm:
## The Problem

Withdraw cancellations throw unhandled error.

## The Solution

Handle.


modules/engine/src/testing/listeners.spec.ts
```diff
@@ -162,15 +169,41 @@ describe(testName, () => {
...
-    it("should should work if responder is bob (DONT submit withdraw to chain)", () => {});
+
+  describe("withdrawal resolution", () => {
+    it("should not create commitments for cancelled withdrawals", async () => {
+      const alice = getRandomChannelSigner();
+      const bob = getRandomChannelSigner();
+      const test = createTestChannelState("create", { alice: alice.address, bob: bob.address });
+      test.transfer.transferState.fee = "0";
+      test.transfer.transferState.nonce = "0";
+      test.transfer.transferState.initiatorSignature = mkSig("0x0");
```
LayneHaber, line 180: This will never be the case, the initiator will always have to supply a valid signature

modules/contracts/src.ts/commitments/withdraw.ts
```diff
@@ -113,16 +113,22 @@ export class WithdrawCommitment {
...
+          recovered = await recoverAddressFromChannelMessage(hash, sig);
+        } catch (e) {
+          recovered = e.message;
+        }
+        if (recovered !== this.alice && recovered !== this.bob) {
+          throw new Error(`Invalid signer detected. Got ${recovered}, expected one of: ${this.alice} / ${this.bob}`);
+        }
+
+        this.aliceSignature = recovered === this.alice ? sig : this.aliceSignature;
+        this.bobSignature = recovered === this.bob ? sig : this.bobSignature;
```
LayneHaber, line 130: I don't like this logic being with the commitment, the commitment (in my view) is something that should always be redeemable onchain. Are we *sure* this needs to be here? Shouldn't cancellations be handled outside of the commitment class

LayneHaber (approved):

PR #541, opened 2021-04-02
package cleanup

sanchaymittal:


rhlsthrm:
Some of these packages are needed. We need `ts-natsutil` in non-dev dependencies.

rhlsthrm (changes requested):
Need to make sure all the packages that need to be deps are deps.

PR #552, opened 2021-04-11
Handle cancelled withdrawals properly

rhlsthrm:

### DIFF
--- a/modules/server-node/src/services/store.ts
+++ b/modules/server-node/src/services/store.ts
@@ -16,10 +16,8 @@ import {
   ChannelDispute,
   TransferDispute,
   GetTransfersFilterOpts,
-  CoreChannelState,
-  CoreTransferState,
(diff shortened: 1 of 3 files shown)


modules/server-node/src/services/store.ts
```diff
@@ -255,7 +253,9 @@ const convertEntitiesToWithdrawalCommitment = (
     bob: channel.participantB,
     recipient: createEntity.transferToA!, // balance = [toA, toB]
     assetId: createEntity.assetId,
-    amount: BigNumber.from(createEntity.transferAmountA).sub(initialState.fee).toString(),
+    amount: BigNumber.from(createEntity.transferAmountA)
+      .sub(initialState.fee ?? 0)
```
LayneHaber, line 257: This should always be defined

LayneHaber (approved):

PR #553, opened 2021-04-11
547 in memory progress

### LINKED ISSUE #547: [router] Make PROCESSING update status an in-memory flag
**Describe the bug**
Router marks queued updates as `PROCESSING` when handling them in the `isAlive` functions. This is not restart-resistant. If a router is upgraded or restarted while an update is being processed, it will never be resumed properly.

The status is designed to prevent errors when people are hammering the `isAlive` endpoint, and this should be handled in-memory instead.

LayneHaber:
## The Problem

Fixes #547

### DIFF
--- a/modules/router/src/forwarding.ts
+++ b/modules/router/src/forwarding.ts
@@ -573,6 +573,12 @@ export async function forwardTransferResolution(
   });
 }
 
+// This function holds updates in memory that are currently
+// being processed. This allows the users to spam the `isAlive`
+// update without the router needing to worry about accidentally
+// double-handling the same checkpointed update
+let inProgressIsAliveChannels: string[] = [];
+
 export async function handleIsAlive(
   data: IsAlivePayload,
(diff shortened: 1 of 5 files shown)


rhlsthrm (approved):

modules/router/src/forwarding.ts
```diff
@@ -1067,10 +1072,12 @@ export const handlePendingUpdates = async (
   }
   const channel = channelRes.getValue() as FullChannelState;
 
+  // This means the user is online and has checked in. Get all updates that are
+  // queued and then execute them.
+  const updates = await store.getQueuedUpdates(data.channelAddress, [RouterUpdateStatus.PENDING]);
```
rhlsthrm, line 1076: try-catch

rhlsthrm (changes requested):
Lgtm pending try catch.

rhlsthrm (approved):

PR #558, opened 2021-04-13
510 eth service tests

rhlsthrm:


modules/types/src/chain.ts
```diff
@@ -72,14 +73,22 @@ export class ChainError extends VectorError {
     BadNonce: "the tx doesn't have the correct nonce",
     InvalidNonce: "Invalid nonce",
     MissingHash: "no transaction hash found in tx response",
-    UnderpricedReplancement: "replacement transaction underpriced",
+    UnderpricedReplacement: "replacement transaction underpriced",
+    AncientBlockSync: "Block information is incomplete while ancient",
```
LayneHaber, line 77: Can we run tests with the direct error strings thrown?

LayneHaber (approved):

PR #560, opened 2021-04-13
Remove `LockService`

### LINKED ISSUE #556: [protocol] Remove `LockService`
See [spec](https://www.notion.so/connext/Remove-LockService-24a31cab9cdf448b8126de3c6aea627c)

LayneHaber:
## The Problem

Fixes #556


modules/engine/src/index.ts
```diff
@@ -1199,15 +1115,15 @@ export class VectorEngine implements IVectorEngine {
...
+      return generateRestoreError(RestoreError.reasons.InvalidSignatures, {
         recoveryError: sigRes.getError().message,
       });
     }
 
     // Verify transfers match merkleRoot
     const { root } = generateMerkleTreeData(activeTransfers);
     if (root !== channel.merkleRoot) {
-      return sendResponseToCounterparty(RestoreError.reasons.InvalidMerkleRoot, {
+      return generateRestoreError(RestoreError.reasons.InvalidMerkleRoot, {
```
That3Percent, line 1126: Everywhere `generateMerkleTreeData` is called needs to call `.free()` on the tree at some point or memory leaks. See above: `const { root } = gen...`. This needs to be `const { root, tree } = gen...; tree.free();`

There are a few cases of this, so please review all instances of `generateMerkleTreeData`

That3Percent, line 1126: The functions `syncDisputes` and `restoreState` are `private` and unused. Either this is a bug or they should be deleted.

modules/engine/src/index.ts
```diff
@@ -1538,6 +1454,18 @@ export class VectorEngine implements IVectorEngine {
     }
   }
 
+  private async runProtocolMethodWithRetries<T = any>(fn: () => Promise<Result<T, ProtocolError>>, retryCount = 5) {
+    let result: Result<T> | undefined;
+    for (let i = 0; i < retryCount; i++) {
+      result = await fn();
+      if (!result.isError) {
```
That3Percent, line 1461: It's a good idea to differentiate between intermittent errors and deterministic errors. That's probably beyond the scope of this PR but worth considering for the future.

That3Percent (changes requested):

modules/engine/src/index.ts
That3Percent: Hmm... actually I think this is fixed in #585 . I'm starting to get confused by the multiple PRs in flight.

LayneHaber: All functions in the engine are marked as private -- you access them via the rpc methods

LayneHaber: Yeah, this PR should be considered the "master PR" that will get merged when we are ready to cut a release. All other PRs are proposed to be merged into this branch specifically. However, this would likely not get fixed by #585 since this is in the engine, and that is for the protocol, so I'll add it in now.

modules/engine/src/index.ts
LayneHaber: Yeah, I agree. We are planning on doing a broader error refactor soon(TM) but for now I think we can retry them all

PR #565, opened 2021-04-15
556 2 queue v0

LayneHaber:
## The Problem

This adds in the rudimentary logic for adding things to the queue, particularly on the receiver side. Should mirror the diagram [here](https://whimsical.com/queue-Y2MgUd8uw9WM3Ds3BUeGEE).

Note that this PR notably does **NOT** include logic around creating the queue. Those functions are stubbed out and may change dramatically in subsequent PRs. The important thing to pay attention to here is the logic in:
- determining if a proposed update should be recreated if one is received at a higher nonce
- logic for determining whether or not to apply a received update


modules/protocol/src/utils.ts
```diff
@@ -162,6 +170,33 @@ export function getParamsFromUpdate<T extends UpdateType = any>(
...
+): FullTransferState {
+  return {
+    balance: update.details.balance,
+    assetId: update.assetId,
+    transferId: update.details.transferId,
+    channelAddress: update.channelAddress,
+    transferDefinition: update.details.transferDefinition,
+    transferEncodings: update.details.transferEncodings,
+    transferTimeout: update.details.transferTimeout,
+    initialStateHash: hashTransferState(update.details.transferInitialState, update.details.transferEncodings[0]),
```
That3Percent, line 185: I get nervous whenever I see these `[0]`. Sometimes they might be correct, like in this case, where I think it refers to Alice & Bob. Sometimes not, like when it's asset ids and is lazy.

Consider these readability improvements for sometime in the future:
- `export type AliceBob<T> = [T, T];`
- Change the type of `transferEncodings` from `string[]` to `AliceBob<string>`
- `export const Alice = 0; export const Bob = 1;`
- `update.details.transferEncodings[Alice]`

If you want to get really fancy you could expose the type to be only allowed to be indexed by tagged by opaque Alice & Bob values, which would prevent any `[0]` without loss of performance because it is still `[0]` under the hood.

Also... is it really correct that this is hardcoded to Alice?

modules/protocol/src/vector.ts
```diff
@@ -34,6 +40,17 @@ export class Vector implements IVectorProtocol {
     [ProtocolEventName.CHANNEL_UPDATE_EVENT]: Evt.create<ChannelUpdateEvent>(),
   };
 
+  // This holds the highest seen nonce (proposed or received) for each channel.
+  // Will be used to determine the priority ordering of the queue currently
+  // being executed
+  private highestNonce: { [channelAddr: string]: number } = {};
```
That3Percent, line 46: TypeScript has a dedicated `Map<Key, Value>` for this sort of thing.

modules/protocol/src/vector.ts
```diff
@@ -136,26 +121,35 @@ export class Vector implements IVectorProtocol {
...
-    const outboundRes = await this.lockedOperation(params);
 
-    return outboundRes;
+    // Pull channel from store
+
+    // Update highest seen nonce to next nonce
+
+    // propose update using sync.outbound
+    // should:
+    // - add params to queue
```
That3Percent, line 131: It's not clear to me that adding params to the queue and sending the proposal should happen in the same method.

I would instead:
Have a method for adding the params to the queue which creates a "bluebird style" manually resolvable promise. The params and the promise can be added to the queue.

Have a separate method which continuously takes works off the queue, executes it, and resolves the promise in a loop. When the queue is empty, it pushes a promise to wake itself up that is resolved when the next item is queued.

In this way, updates will be serialized as if there is a lock. There cannot be race conditions by updating stale data.

I'd be happy to write this queue for you if you think it's the right abstraction.

That3Percent (commented):
Here's part of the review. Still need to read more but pushing the thoughts so far.

modules/protocol/src/utils.ts
LayneHaber: No this is specifically for `transferEncodings` which are returned from the `RegisteredTransfer` information onchain. The `transferEncodings` are always ordered `[stateEncoding, resolverEncoding]`.

We could still implement these changes for the state/resolver encodings, though

modules/protocol/src/vector.ts
LayneHaber: Are you talking about using a `Map` class (with the `get`, `set`, `has`) interface or using just the `Map<K, V>` type? If it's just the type I personally prefer the typing above since then you get more context for what the `K` type should be instead of just `string`

modules/protocol/src/vector.ts
LayneHaber: The nuance here (that I punted on mostly) is when you are out of sync. 

Currently when you are out of sync you have to apply the update that you dropped, then regenerate and resend the params. This worked well in the lock-based system, but I think for the nonce based system your way of doing things is cleaner. The concern I have about the out of sync case is you have to adopt and validate your signature on an update that has a stale nonce, which may be a bit odd for the other "only higher nonces" assumption. For deposits you can mostly ignore things since it's a symmetric update, but for `create` and `resolve` updates the transfer in your store will need to be created/updated.

I think we'll have to break up the `outbound` method -- it exists as all one thing since it uses stored data to generate the `nonce` when creating the update (sent over the wire) from the params (supplied to the API).

modules/protocol/src/vector.ts
```diff
@@ -332,6 +393,97 @@ export class Vector implements IVectorProtocol {
...
+          new QueuedUpdateError(QueuedUpdateError.reasons.MissingTransferForUpdateInclusion, {
+            proposedUpdate,
+            receivedUpdate,
+          }),
+        );
+      }
+      const { tree, root, proof } = generateMerkleTreeData(updatedActiveTransfers, proposedTransfer);
+      const included = tree.verify(proof!, bufferify(hashCoreTransferState(proposedTransfer)), root);
+      return Result.ok(included);
+    }
```
LayneHaber, line 480: This will likely also always return `false` since the `transfer.initiator` and `transfer.responder` would not be the same (unless you received an update where your counterparty created a transfer with the `transfer.initiator` as you, which should fail validation).

Probably worth:
1. Making sure the case where your counterparty proposes an update with you as the `transfer.initiator` fails
2. Removing this to reduce any latency caused by regenerating the merkle tree unnecessarily

modules/protocol/src/vector.ts
LayneHaber: But yes, I think if you want to start building that that would be great. I'll be traveling all day on tomorrow, but happy to do some work on it over the weekend if you start it on a separate branch and leave some notes on where to pick it up from in the comments

modules/protocol/src/vector.ts
```diff
@@ -263,50 +257,117 @@ export class Vector implements IVectorProtocol {
...
+        // case you should retry the proposed update anyway (errors
+        // should fall through to validation)
+
+        // If it does not, insert previously in progress update
+        // at the front of queue
+        const regenerated = getParamsFromUpdate(proposed);
+        if (regenerated.isError) {
+          return;
+        }
+        const processedResult = await this.addToFrontOfQueue(regenerated.getValue(), received.update);
```
That3Percent, line 365: This has me a little worried that it's not just taking the same path. (Eg: At the top of the method this uses `addToQueue` instead of `addToFrontOfQueue`. The proposed update is just the same now as a normal update that would have been added later, so a more natural way to do this in a lock-free system would be to put a `while (true)` at the top of this method so that everything retries including nonce setting, etc. `while(true)` is a very common pattern to see in a lock-free construction.

In addition to the above, I find a lot of this hard to read but am hoping that implementing the queue or something to serialize access to data will tame this. I've got a pretty good idea how to approach the queue without the sync protocol and disputes... but a queue without those isn't good enough.

Getting this down to "obviously correct" is going to be a struggle but that might be part of the territory.

modules/protocol/src/vector.ts
```diff
@@ -332,6 +393,97 @@ export class Vector implements IVectorProtocol {
...
+        updatedTransfer?: FullTransferState;
+      },
+      QueuedUpdateError
+    >
+  > {
+    throw new Error("addToQueue method not implemented");
+  }
+
+  /**
+   * Returns true if the received upddate includes our proposed update. If true,
```
That3Percent, line 433: Typo: "upddate" -> "update"

modules/protocol/src/vector.ts
```diff
@@ -263,50 +257,117 @@ export class Vector implements IVectorProtocol {
...
+
+        // If no, return
+        if (!proposed) {
+          return;
+        }
+
+        // If so, check if the update includes our proposed update
+        const proposedTransfer =
+          proposed.type === UpdateType.create ? getTransferFromUpdate(proposed, updatedChannel) : undefined;
+        const included = this.includesOurProposedUpdate(
```
That3Percent, line 333: There may be a wrinkle here to consider.

In this scenario where the proposed update is included we have to realize that we may not have the double-signed state for our proposed update. What if their proposed update is rejected? Do the participants disagree on the latest double-signed state?

modules/protocol/src/vector.ts
```diff
@@ -332,6 +393,97 @@ export class Vector implements IVectorProtocol {
...
+    updatedActiveTransfers: FullTransferState[],
+    proposedTransfer?: FullTransferState,
+  ): Result<boolean, QueuedUpdateError> {
+    // If both are a setup update, your update would fail and the ultimate
+    // result (a channel is set up) is achieved
+    if (receivedUpdate.type === UpdateType.setup && proposedUpdate.type === UpdateType.setup) {
+      return Result.ok(true);
+    }
+
+    // If both are a deposit update, your deposit is implicitly included
```
That3Percent, line 453: Your deposit is implicitly included? I'll have to take your word for it.

modules/protocol/src/vector.ts
```diff
@@ -332,6 +393,97 @@ export class Vector implements IVectorProtocol {
...
+    if (receivedUpdate.type === UpdateType.deposit && proposedUpdate.type === UpdateType.deposit) {
+      return Result.ok(true);
+    }
+
+    // If both are a resolve, it would *not* include your proposed update since
+    // only the responder to a transfer can call resolve. Updates by definition
+    // are different
```
That3Percent, line 460: I don't follow here.

The reason that the proposed update could include your resolve is that you just sent them the resolve.

The thing we're checking for here generally is that they received your update, processed it, sent a response, then processed their update and sent an update but for whatever reason the first response was dropped or is received out of order.

In the scenario above it would include your proposed update.

modules/protocol/src/vector.ts
```diff
@@ -332,6 +393,97 @@ export class Vector implements IVectorProtocol {
...
+    if (receivedUpdate.type === UpdateType.resolve && proposedUpdate.type === UpdateType.resolve) {
+      return Result.ok(false);
+    }
+
+    // If both are a create, it would include your proposed update IFF the
+    // merkle root *after* the update was applied includes the transfer you
+    // attempted to create
+    if (receivedUpdate.type === UpdateType.create && proposedUpdate.type === UpdateType.create) {
```
That3Percent, line 468: The first check to see if both are setup makes sense, but all these branches after I'm a bit lost.

The way I see it there's 3 possibilities and they are independent of the types of updates.
  1. The final state is (initial + our update + their update) => return true
  2. The final state is (initial + their update) => return false
  3.  The final state is something else => fail
This can be detected by just playing the updates forward and seeing what the merkle root is.

That3Percent (changes requested):

modules/protocol/src/utils.ts
That3Percent: Yeah, the fact that I guessed wrong is a datapoint indicating that readability could be improved.

modules/protocol/src/vector.ts
That3Percent: Ah. So you like that it is self documenting there. Another way to approach that is have `K, V` use tagged types. Eg: `Map<ChannelAddr, Nonce>`. The problem with objects generally is that you can shadow properties, and there are performance implications. Those are probably not a huge deal here (the chances that an address will be the name of a property is 0), but sometimes it's a security risk.

Jannis: There's also `Record<Key, Value>`, which may be more appropriate.

modules/protocol/src/vector.ts
LayneHaber: You are right

modules/protocol/src/vector.ts
```diff
@@ -332,6 +393,97 @@ export class Vector implements IVectorProtocol {
...
+    receivedUpdate: ChannelUpdate,
+    proposedUpdate: ChannelUpdate,
+    // TODO: should probably just use the generated merkle tree here instead
+    // of regenerating it from the transfers and channel
+    updatedActiveTransfers: FullTransferState[],
+    proposedTransfer?: FullTransferState,
+  ): Result<boolean, QueuedUpdateError> {
+    // If both are a setup update, your update would fail and the ultimate
+    // result (a channel is set up) is achieved
+    if (receivedUpdate.type === UpdateType.setup && proposedUpdate.type === UpdateType.setup) {
```
That3Percent, line 449: After some further consideration I think this specific case to be invalid. Both participants cannot be Alice in the contract that was actually deployed on-chain. Only Alice should send update 0.

That leaves only the "try two" method for all cases, where if the wrong participant performs setup it should fail for that validation.

That3Percent (changes requested):

LayneHaber:
Closing, will go different route

PR #568, opened 2021-04-15
Dockerfile(s): change curl command to enable multiple retries of curl command

jakekidd:
## The Problem
You may have seen this issue (during build of vector with `make`) flying around lately:
```
#9 55.74   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#9 55.74                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0curl: (6) Could not resolve host: raw.githubusercontent.com
------
executor failed running [/bin/sh -c apk add --no-cache bash curl g++ gcc git jq make python2 &&    npm config set unsafe-perm true &&    npm install -g lerna@$LERNA_VERSION &&    curl https://raw.githubusercontent.com/vishnubob/wait-for-it/$(git ls-remote https://github.com/vishnubob/wait-for-it.git refs/heads/master | cut -f1)/wait-for-it.sh > /bin/wait-for &&    chmod +x /bin/wait-for &&    chmod +x /*.sh &&    rm -rf /var/cache/apk/* /tmp/*]: exit code: 6
make: *** [Makefile:265: builder] Error 1
```
Various solutions were 'found' - but no one solution seemed to be settled upon:
- update /etc/docker/daemon.json to have "dns": ["8.8.8.8"]
- for WSL2: update resolv.conf (after setting it to *not* be generated in wsl.conf) to set DNS to 8.8.8.8 manually
- upgrade docker (I did this oops)
- various combinations of updating docker and restarting computer, etc.
Personally, nothing above worked for me until I made the change in this PR.

(FYI: I am on Windows 10 using WSL2 for Ubuntu.)

## The Solution
Ultimately this is a sort of 'brute force' fix where we just retry the curl command.

The DNS issue was happening because the default DNS was unable to resolve `raw.githubusercontent.com` - however, there may be a 'transitional' problem here where it does not retry (or even try at all) using the backup DNS (8.8.8.8 in this case).

**Potential Issue with this Solution**:
- I used `until`, so basically this will keep going until it resolves. It may not resolve if you do not have backup DNS set (8.8.8.8)... or if your internet randomly sputters out, so the user would have to ^C out of the `make` process.

May need to do a set number of retries instead of `until`.

### DIFF
--- a/modules/auth/ops/Dockerfile
+++ b/modules/auth/ops/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /app
 
 RUN apk add --no-cache bash curl g++ gcc git jq make python2 &&\
     npm config set unsafe-perm true &&\
-    curl https://raw.githubusercontent.com/vishnubob/wait-for-it/$(git ls-remote https://github.com/vishnubob/wait-for-it.git refs/heads/master | cut -f1)/wait-for-it.sh > /bin/wait-for &&\
+    until curl https://raw.githubusercontent.com/vishnubob/wait-for-it/$(git ls-remote https://github.com/vishnubob/wait-for-it.git refs/heads/master | cut -f1)/wait-for-it.sh > /bin/wait-for; do sleep 5; done &&\
     chmod +x /bin/wait-for &&\
     rm -rf /var/cache/apk/* /tmp/*
 
--- a/modules/contracts/ops/Dockerfile
+++ b/modules/contracts/ops/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /app
 
 RUN apk add --no-cache bash curl g++ gcc git jq make python2 &&\
     npm config set unsafe-perm true &&\
-    curl https://raw.githubusercontent.com/vishnubob/wait-for-it/$(git ls-remote https://github.com/vishnubob/wait-for-it.git refs/heads/master | cut -f1)/wait-for-it.sh > /bin/wait-for &&\
+    until curl https://raw.githubusercontent.com/vishnubob/wait-for-it/$(git ls-remote https://github.com/vishnubob/wait-for-it.git refs/heads/master | cut -f1)/wait-for-it.sh > /bin/wait-for; do sleep 5; done &&\
     chmod +x /bin/wait-for &&\
     rm -rf /var/cache/apk/* /tmp/*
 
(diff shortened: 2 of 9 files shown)


rhlsthrm:
Looks good, I want to get @sfoxdev to look at it too though.

LayneHaber (approved):
lgtm but would like a review from @sfoxdev as well

sfoxdev:
@jakekidd such kind of DNS issues you should resolve only on your local machine. We can't add `until` command for all docker images as we use CI for automated builds so it could slow down them or make some loop trap.

jakekidd:
> @jakekidd such kind of DNS issues you should resolve only on your local machine. We can't add `until` command for all docker images as we use CI for automated builds so it could slow down them or make some loop trap.

Alright. I've exhausted every solution on that list - every DNS option I could find (Docker is set manually to use 8.8.8.8). Everything I've done has point me towards this being some kind of transitional problem in the build process (a buried race condition, or the DNS just isn't changing).

Additionally, this error isn't just localized to my machine, many others have repro'd.

So going forward, just making a note that I'm literally going to have to keep this change on my local computer to continue building.

@sfoxdev any chance you'd be open to maybe one retry of this line? Assuming your concern here is just that we could get caught in loop trap. This solution could be unique for WSL2/Windows, but we should be compatible with that stack in general

sfoxdev:
@jakekidd To reproduce your problem I get new laptop with fresh installed Windows 10 and install Docker and WSL2 on it. Then I built all images without any curl issue. DNS works correctly and resolve all domains. 
This error is definitely localized to your machine. If anyone else have the same issue then comment here with step-by-step guide how to reproduce it.

From my practice working with docker when you faced with DNS issue then you need to find and solve problem locally.

jakekidd:
> @jakekidd To reproduce your problem I got new laptop with fresh installed Windows 10 and install Docker and WSL2 on it. Then I built all images without any curl issue. DNS works correctly and resolve all domains.
> This error is definitely localized to your machine. If anyone else have the same issue then comment here with step-by-step guide how to reproduce it.
> 
> From my practice working with docker when you faced with DNS issue then you need to find and solve problem locally.

Okay. Any insight as to why it would work on the second curl request and not the first?

I spent days on this, couldn't resolve why my machine was having the issue. @LayneHaber and @rhlsthrm have also encountered the issue, can't recall how they repro'd it, but they did so after a couple clients experienced the same thing. Clients resolved it with varying solutions (see my PR opening comment)

If your windows 10 / docker / wsl2 setup didn't include any additional modifications (like explicit DNS in docker, change to daemon.json, etc.) then I'm pretty mystified here. (Please lmk if you did do anything additional here)

PR #574, opened 2021-04-20
Amm

rhlsthrm:


modules/router/src/services/config.ts
```diff
@@ -81,6 +85,93 @@ export const getSwapFees = (
...
+          routerSignerAddress,
+        }),
+      );
+    }
+    balances.push(onChainRouterBalance);
+  }
+
+  const transferAmountBn = BigNumber.from(transferAmount);
+  const stableAmmAddress = getConfig().stableAmmAddress!;
+  const stableAmmProvider: JsonRpcProvider = new JsonRpcProvider(getConfig().stableAmmProvider!);
```
jakekidd, line 147: Are these force unwraps "safe" here? As in, if the user does not define a valid env var for `STABLE_AMM_PROVIDER`, and calls this method `onSwapGivenIn`, is this going to error out in a way that's useful?

Or do we know for a fact they could not have gotten this far without those vars defined?

rhlsthrm: Yeah, this is handled at the config layer itself.

modules/router/src/test/services/config.spec.ts
```diff
@@ -23,4 +45,70 @@ describe("config.ts", () => {
...
+      const estimatedPrice = parseEther("1");
+      const res = getPriceImpact(marketPrice, estimatedPrice);
+      console.log(res);
+    });
+
+    it("happy: getMappedAssets 1:1.2", () => {
+      const marketPrice = parseEther("1");
+      const estimatedPrice = parseEther("1.2");
+      const res = getPriceImpact(marketPrice, estimatedPrice);
+      console.log(res.toString());
```
jakekidd, line 72: I think we should actually ensure desired behavior here for above three methods, no? If this is just a TODO item, I'll take care of it

modules/router/src/test/services/config.spec.ts
```diff
@@ -23,4 +45,70 @@ describe("config.ts", () => {
...
+        toAssetId,
+        toChainId,
+        routerSignerAddress,
+        ethReader as IVectorChainReader,
+        log,
+      );
+
+      expect(res.isError).to.be.true;
+      expect(res.getError()!.message).to.be.eq(ConfigServiceError.reasons.UnableToGetSwapRate);
+    });
```
jakekidd, line 110: Think we need one more test here to check for desired behavior in a valid use case. I will add shortly

modules/router/src/services/config.ts
```diff
@@ -81,6 +97,157 @@ export const getSwapFees = (
...
+    }
+  });
+
+  const uniquePairs = Array.from(new Set(pair.map((s) => s.assetId))).map((x) => {
+    return pair.find((z) => z.assetId === x)!;
+  });
+  return uniquePairs;
+};
+
+export const getPriceImpact = (marketPrice: BigNumber, estimatedPrice: BigNumber): BigNumber => {
```
jakekidd, line 118: Do we return price impact as a percentage integer, as opposed to a decimal? Seems a bit dangerous - we are assuming method callers/users will know it's returning a perc integer.

I'm adding a method header just to clarify for now. We should be doing better docstrings soon, so this might not be worth concern rn

jakekidd: Also, I thought price impact was only for effect of a token swap? Not diff between estimated price and market price in perc

jakekidd: Math should be:
```
// Assuming user is swapping token B for (-->) token A in router:
// Get token's current prices.
priceTokenA = ..
priceTokenB = ..
// Get amounts for each token after executing the desired swap.
amountTokenA = <amount of token A in liquidity pool> - <amount being withdrawn>
amountTokenB = <amount of token B in liquidity pool> + <amount being deposited>
// Calculate new projected price of token after execution.
projectedPriceTokenB = amountTokenA * priceTokenA / amountTokenB
// Convert to percentage difference from previous value.
1 - projectedPriceTokenB / priceTokenB
```

jakekidd: Alternatively, I could be misunderstanding this method's purpose, and it could just be a case of a misnomer here

jakekidd: If you wanted to account for fees, assuming we take those in tokenA, we would change this one line:
```
// Assuming a 3% fee:
projectedPriceTokenB = 0.97 * amountTokenA * priceTokenA / amountTokenB
```

sanchaymittal (approved):

PR #576, opened 2021-04-21
Routing complete

### LINKED ISSUE #575: [node] Add ROUTING_COMPLETE event

rhlsthrm:
## The Problem

Fixes #575

### DIFF
--- a/modules/engine/src/listeners.ts
+++ b/modules/engine/src/listeners.ts
@@ -36,6 +36,7 @@ import {
   IVectorStore,
   DEFAULT_FEE_EXPIRY,
   DEFAULT_CHANNEL_TIMEOUT,
+  RouterSchemas,
 } from "@connext/vector-types";
 import { getRandomBytes32, hashWithdrawalQuote, mkSig } from "@connext/vector-utils";
 import { getAddress } from "@ethersproject/address";
@@ -46,7 +47,6 @@ import Pino, { BaseLogger } from "pino";
 import { IsAliveError, RestoreError, WithdrawQuoteError } from "./errors";
 
 import { EngineEvtContainer } from "./index";
-import { normalizeGasFees } from "./utils";
 
 export async function setupEngineListeners(
   evts: EngineEvtContainer,
@@ -95,7 +95,17 @@ export async function setupEngineListeners(
   // Set up listener for conditional transfer creations
   vector.on(
     ProtocolEventName.CHANNEL_UPDATE_EVENT,
-    async (event) => await handleConditionalTransferCreation(event, store, chainService, chainAddresses, evts, logger),
+    async (event) =>
+      await handleConditionalTransferCreation(
+        event,
+        store,
+        chainService,
+        chainAddresses,
+        evts,
+        logger,
+        signer,
+        messaging,
+      ),
     (event) => {
       const {
         updatedChannelState: {
@@ -405,6 +415,14 @@ export async function setupEngineListeners(
     logger.info({ quote: calculatedQuote.toJson(), method, methodId }, "Method complete");
   });
 
+  // if this message is received, means that a transfer we sent was routed completely to the end recipient
+  // emit the appropriate event to our external listeners
+  await messaging.onReceiveTransferRoutingCompleteMessage(signer.publicIdentifier, async (data) => {
+    evts.CONDITIONAL_TRANSFER_ROUTING_COMPLETE.post(data.getValue());
+
+    // in a multihop world we would check if we routed this transfer ourselves and if so we would
+  });
+
   ////////////////////////////
   /// CHAIN SERVICE EVENTS
   chainService.on(EngineEvents.TRANSACTION_SUBMITTED, (data) => {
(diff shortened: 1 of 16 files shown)


modules/test-runner/src/trio/happy.test.ts
```diff
@@ -153,7 +152,7 @@ describe(testName, () => {
     );
   });
 
-  it("ETH: deposit, transfer C -> R -> D, withdraw", async () => {
+  it.only("ETH: deposit, transfer C -> R -> D, withdraw", async () => {
```
jakekidd, line 155: .only, is this okay here? orrr

modules/engine/src/listeners.ts
```diff
@@ -405,6 +415,14 @@ export async function setupEngineListeners(
     logger.info({ quote: calculatedQuote.toJson(), method, methodId }, "Method complete");
   });
 
+  // if this message is received, means that a transfer we sent was routed completely to the end recipient
+  // emit the appropriate event to our external listeners
+  await messaging.onReceiveTransferRoutingCompleteMessage(signer.publicIdentifier, async (data) => {
+    evts.CONDITIONAL_TRANSFER_ROUTING_COMPLETE.post(data.getValue());
+
+    // in a multihop world we would check if we routed this transfer ourselves and if so we would
```
jakekidd, line 423: just double checking - should there be a TODO item here?

modules/engine/src/listeners.ts
```diff
@@ -650,10 +670,42 @@ async function handleConditionalTransferCreation(
   };
   evts[EngineEvents.CONDITIONAL_TRANSFER_CREATED].post(payload);
 
-  // If we should not route the transfer, do nothing
-  if (!routingId || transfer.meta?.routingId !== routingId) {
-    logger.warn({ transferId, routingId, meta: transfer.meta }, "Cannot route transfer");
-    return;
+  // TRANSFER_ROUTING_COMPLETE event
+
+  // if transfer is created and we are the end recipient, emit event now and end
```
jakekidd, line 675: Another double check - 'emit event now and end' - should we be returning in this block (based on that phrasing)?

modules/test-runner/src/trio/happy.test.ts
rhlsthrm: No it is def not. That was for testing.

modules/engine/src/listeners.ts
rhlsthrm: That's correct, fixed.

modules/engine/src/testing/listeners.spec.ts
```diff
@@ -741,4 +745,127 @@ describe(testName, () => {
...
+          name: "foo",
+          encodedCancel: "0x",
+          resolverEncoding: "bar",
+          stateEncoding: "baz",
+        }),
+      );
+    });
+
+    it("should emit CONDITIONAL_TRANSFER_ROUTING_COMPLETE for the end recipient", async () => {
+      return new Promise((res, rej) => {
```
LayneHaber, line 764: Will returning a promise in a test actually properly work here? can you `await` it instead?

modules/engine/src/testing/listeners.spec.ts
```diff
@@ -741,4 +745,127 @@ describe(testName, () => {
...
+      ]);
+
+      evts.CONDITIONAL_TRANSFER_ROUTING_COMPLETE.attachOnce((data) => {
+        expect(data).to.deep.eq({
+          initiatorIdentifier: test.transfer.initiatorIdentifier,
+          responderIdentifier: test.transfer.responderIdentifier,
+          routingId: test.transfer.meta.routingId,
+          meta: test.transfer.meta,
+        } as ConditionalTransferRoutingCompletePayload);
+      });
```
LayneHaber, line 845: How can you ensure the event was emitted properly and the structure is asserted if this isn't a `waitFor`? It could be that the other expects succeed, and the test suite is terminated.

modules/test-runner/src/trio/happy.test.ts
```diff
@@ -171,25 +170,25 @@ describe(testName, () => {
...
+    // );
+    // // bob withdraws
+    // await withdraw(
+    //   daveService,
+    //   daveRogerPostSetup.channelAddress,
+    //   assetId,
+    //   withdrawAmt.sub(transferQuote.fee).div(2),
+    //   Wallet.createRandom().address,
+    //   true,
+    // );
```
LayneHaber, line 191: Is this still supposed to be commented out?

LayneHaber (changes requested):

modules/engine/src/testing/listeners.spec.ts
rhlsthrm: This is the documented way to do it in Mocha. It's working also.

LayneHaber: Huh neat-o

LayneHaber (approved):

modules/engine/src/listeners.ts
```diff
@@ -650,11 +672,56 @@ async function handleConditionalTransferCreation(
...
+      responderIdentifier: transfer.responderIdentifier,
+      routingId,
+      meta: transfer.meta,
+    });
     return;
   }
+
+  // if we are the exit router, also send out-of-protocol NATS message to the sender
+  // of the linked transfer so they know the path is complete (and can forward to their sender and so on)
+  if (
```
jakekidd, line 694: nit: alternative to return above, you could just make this an `else if`, that way if there's any more NATS messages we need to send here based on other conditions in the future, no room for mistake

but that's nit. otherwise lgtm 👍

jakekidd (approved):
left one last nit but not necessary, lgtm

PR #580, opened 2021-04-23
Switch merkle libs

LayneHaber:
## The Problem

Switches the merkle tree lib to the one from @That3Percent


modules/utils/src/merkle.ts
```diff
@@ -15,7 +16,9 @@ export const generateMerkleTreeData = (transfers: CoreTransferState[]): { root:
   });
 
   // Generate tree
-  const tree = new MerkleTree(leaves, keccak256, { sortPairs: true });
+  const tree = new merkle.Tree();
```
That3Percent, line 19: This isn't well documented in `wasm_bindgen`, but you'll need to call `free` on the tree or this will leak memory.

modules/utils/src/merkle.ts
```diff
@@ -15,7 +16,9 @@ export const generateMerkleTreeData = (transfers: CoreTransferState[]): { root:
...
+  tree.insert_hex_js(leaves);
```
That3Percent, line 20: This method takes just one leaf at a time. So, you'll need to loop over them. Remember, the intent was that the tree is updated iteratively (though the code you have written is a shorter and less risky path - I haven't profiled to confirm that this will actually be better if we don't keep the tree around. It's worth verifying that this is an improvement over the old code since this is not what was profiled when reporting the improvements.)

That3Percent, line 20: `insert_hex_js` expects a string from `encodeCoreTransferState`, not `hashCoreTransferState`. It decodes the `transferId` from there, and does it's own hashing.

That3Percent, line 20: The Rust implementation does it's own sorting. So, you can get rid of the sorting by `transferId` above.

That3Percent (changes requested):

modules/utils/src/merkle.ts
```diff
@@ -1,26 +1,36 @@
...
-  // Sort transfers alphabetically by id
-  const sorted = transfers.sort((a, b) => a.transferId.localeCompare(b.transferId));
+import { encodeCoreTransferState, hashCoreTransferState } from "./transfers";
 
+export const generateMerkleTreeData = (transfers: CoreTransferState[]): { root: string; tree: merkle.Tree } => {
   // Create leaves
-  const leaves = sorted.map((transfer) => {
-    return hashCoreTransferState(transfer);
+  const tree = new merkle.Tree();
+  tree.free(); // handle memory leaks
```
That3Percent, line 12: You need to call `free` after you're done using the Tree, not right after construction.

Example:
```typescript
const tree = new merkle.Tree();
let calculated;
try {
   // TODO: Insert hex here
   calculated = tree.root_js();
} finally {
   tree.free();
}

File: modules/utils/src/merkle.ts [hunk 4]
```diff
@@ -1,26 +1,36 @@
...
+    tree.insert_hex_js(encodeCoreTransferState(transfer));
   });
 
-  // Generate tree
-  const tree = new MerkleTree(leaves, keccak256, { sortPairs: true });
-
   // Return
-  const calculated = tree.getHexRoot();
+  const calculated = tree.root_js();
+
```
That3Percent, line 19: Do not return the `tree` object once it has been `free`'d. Any method call will throw an exception.

That3Percent (changes requested):

modules/utils/src/merkle.ts
```diff
@@ -9,13 +9,17 @@ import { encodeCoreTransferState, hashCoreTransferState } from "./transfers";
...
-  tree.free(); // handle memory leaks
   transfers.forEach((transfer) => {
     tree.insert_hex_js(encodeCoreTransferState(transfer));
   });
 
   // Return
-  const calculated = tree.root_js();
+  let calculated: string;
+  try {
+    calculated = tree.root_js();
```
That3Percent, line 19: The part where the transfers are inserted needs to be in the body of the `try` block.

modules/utils/src/merkle.ts
```diff
@@ -9,13 +9,17 @@ import { encodeCoreTransferState, hashCoreTransferState } from "./transfers";
...
+  } finally {
+    tree.free(); // handle memory leaks
```
That3Percent, line 21: We don't usually say "handle memory leaks" every time memory is `free`'d.

That3Percent (changes requested):

modules/utils/src/merkle.spec.ts
```diff
@@ -9,7 +9,7 @@ import { keccak256 } from "ethereumjs-util";
 import { keccak256 as solidityKeccak256 } from "@ethersproject/solidity";
 import { bufferify } from "./crypto";
 
-describe("generateMerkleTreeData", () => {
+describe.only("generateMerkleTreeData", () => {
```
That3Percent, line 12: This should not be `only`

modules/utils/src/merkle.ts
```diff
@@ -1,26 +1,40 @@
...
 import { MerkleTree } from "merkletreejs";
 
-import { hashCoreTransferState } from "./transfers";
-
-export const generateMerkleTreeData = (transfers: CoreTransferState[]): { root: string; tree: MerkleTree } => {
-  // Sort transfers alphabetically by id
-  const sorted = transfers.sort((a, b) => a.transferId.localeCompare(b.transferId));
+import { encodeCoreTransferState, hashCoreTransferState } from "./transfers";
 
+export const generateMerkleTreeData = (transfers: CoreTransferState[]): { root: string; tree: merkle.Tree } => {
```
That3Percent, line 9: Looks like that previous comment about not returning an unusable `merkle.Tree` (that has been `free`'d) has not yet been implemented.

That3Percent (changes requested):

modules/utils/src/merkle.spec.ts
```diff
@@ -18,17 +18,29 @@ describe("generateMerkleTreeData", () => {
       });
   };
 
-  it("should work for a single transfer", () => {
+  // TODO: proof and verification on new tree
+  it.skip("should work for a single transfer", () => {
```
That3Percent, line 22: And these should not be `.skip` or have other important parts of the test commented out. (There's a few cases of this here, I won't comment on each)

That3Percent (changes requested):

modules/utils/src/merkle.ts
```diff
@@ -1,26 +1,40 @@
...
+  } finally {
+    tree.free(); // handle memory leaks
+  }
+
   return {
     root: calculated === "0x" ? HashZero : calculated,
     tree,
   };
 };
+
```
That3Percent, line 29: I didn't notice this above because it's not part of the review, but this `root: calculated === "0x" ? HashZero : calculated,` can just be `root: calculated` because the lib already returns `HashZero`. With this simplification, you might as well also rename `calculated` to `root` and then have `return { root }`, which could just become `return root` after removing `tree`.

That3Percent (changes requested):

PR #585, opened 2021-04-26
556 switch lib

LayneHaber:
## The Problem

- Uses more efficient merkle root generation by adding/removing leaves as needed
- Switches to new merkle lib


modules/protocol/src/errors.ts
```diff
@@ -66,6 +66,7 @@ export class CreateUpdateError extends ProtocolError {
     CouldNotSign: "Failed to sign updated channel hash",
     FailedToReconcileDeposit: "Could not reconcile deposit",
     FailedToResolveTransferOnchain: "Could not resolve transfer onchain",
+    FailedToUpdateMerkleRoot: "Could not generate new merkle root",
```
That3Percent, line 69: This seems really unlikely. The only reason this should ever throw is if there is a bug.

modules/protocol/src/vector.ts
```diff
@@ -39,6 +46,9 @@ export class Vector implements IVectorProtocol {
   // Hold the serialized queue for each channel
   private queues: Map<string, SerializedQueue<SelfUpdateResult, OtherUpdateResult>> = new Map();
 
+  // Hold the merkle tree for each channel
+  private trees: Map<string, merkle.Tree> = new Map();
```
Jannis, line 50: How much memory will this realistically consume? It could make sense to prune entries after not being used for a while or based on a maximum number of trees to keep around. A library I can recommend is https://www.npmjs.com/package/@thi.ng/cache, which provides both an LRU cache as well as a TLRU cache.

LayneHaber: Hmm good question, let me check out that library. Thanks for the tip!

modules/utils/src/merkle.ts
```diff
@@ -49,7 +49,7 @@ export const addTransferToTree = (transfer: CoreTransferState, tree: merkle.Tree
 export const removeTransferFromTree = (transfer: CoreTransferState, tree: merkle.Tree): MerkleTreeUpdate => {
   let root: string;
   try {
-    tree.insert_hex_js(encodeCoreTransferState(transfer));
+    tree.delete_id_js(transfer.transferId);
```
That3Percent, line 52: This and other places are still calling `tree.free` at the wrong time.

Remember: You cannot call any method on `tree` after calling `free`.

The right time to call `free` would be after calling `this.trees.delete(channelAddress)`. Since this doesn't seem to happen, just don't call free at all.

That3Percent (changes requested):

modules/protocol/src/errors.ts
LayneHaber: Yeah the issue is more that *if* it fails due to a bug it shouldn't throw an error it should return a `Result.fail(...)` so it can be handled properly at the higher levels

modules/protocol/src/vector.ts
```diff
@@ -493,6 +505,34 @@ export class Vector implements IVectorProtocol {
...
+    activeTransfers: FullTransferState[],
+    transfer: FullTransferState,
+    update: typeof UpdateType.create | typeof UpdateType.resolve,
+  ): Result<string> {
+    if (!this.trees.has(channelAddress)) {
+      console.log("***** generating new merkle tree data");
+      const { tree } = generateMerkleTreeData(activeTransfers);
+      this.trees.set(channelAddress, tree);
+    } else {
+      console.log("***** updating existing tree, yay!");
```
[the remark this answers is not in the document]
LayneHaber: I do not :)

modules/protocol/src/vector.ts
That3Percent: This will consume approximately 64 bytes per transfer (lower bound, not counting sizing Vec buffer). So, if the router has a million transfers this would be a "whopping" 64 megabytes. I think this will be ok and that adding something here like an LRU cache isn't worth the complexity/effort/performance. Most LRU caches that I've seen in the wild just make performance worse.

modules/protocol/src/sync.ts
```diff
@@ -264,31 +264,19 @@ export async function inbound(
   // (d,e) update.nonce > expectedInSync -- restore case handled in syncState
 
   // Get the difference between the stored and received nonces
-  const channelNonce = channel?.nonce ?? 0;
   const ourPreviousNonce = channel?.latestUpdate?.nonce ?? -1;
-  const aliceSentUpdate = update.type === UpdateType.setup ? true : update.fromIdentifier === channel?.aliceIdentifier;
 
-  // Get the expected nonce
-  const expectedNonce = getNextNonceForUpdate(channelNonce, aliceSentUpdate);
+  // Get the expected previous update nonce
```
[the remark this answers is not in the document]
LayneHaber: The `??` only operates on `null` or `undefined` values

PR #588, opened 2021-04-27
Tx Retry Fixes

jakekidd:
## The Problem
Copied from: https://www.notion.so/connext/Withdrawal-Retry-Fixes-d355810cb2944818ae63b601bac04f85

Withdrawals in some cases will not be fully submitted. These are some of the failure cases we have recently experienced:

- Withdrawal commitment never submitted onchain. Could be from ill-timed router restart, or provider failure.
- Withdrawal commitment submitted onchain, tx hash is available, but tx receipt is not available. This is effectively a "bad tx" that needs to be retried with the same parameters.
- Withdrawal commitment is submitted on chain, but we have no record of its receipt. Can be caused by ill-timed router restart while waiting for the receipt.

## The Solution
From: https://www.notion.so/connext/Withdrawal-Retry-Fixes-d355810cb2944818ae63b601bac04f85

- [x]  Submit tx to chain. Catch retryable errors and retry if failure. Ex: Nonce too low means there's a race with other txs sending, so keep retrying until it works. Record the following in DB: `nonce`, `to`, `value`, `data`, `hash`, `gasPrice`, `gasLimit`. Emit `TRANSACTION_SUBMITTED` event.
- [ ]  Poll every 2 seconds until `getTransactionReceipt` returns a valid receipt. After 15 seconds, speed up the tx by sending at the same nonce with higher gas price. Amount to bump price can be configurable.
    - Note: Do not trust Ethers listeners for this process since the `tx.wait()` might resolve but no tx receipt will be available, especially for non-ETH chains.
- [x]  Save tx receipt to DB. If `receipt.status === 0`, tx is reverted, Emit `TRANSACTION_REVERTED` event, else emit `TRANSACTION_MINED` event.
- [ ]  On node startup:
    - Pull transactions that have a tx hash but no receipt.
    - Check if txs have been mined. If so, save receipt to DB. If `receipt.status === 0`, tx is reverted, Emit `TRANSACTION_REVERTED` event, else emit `TRANSACTION_MINED` event.
    - If not, resubmit at same nonce and higher gas price. (TODO: How to handle nonce errors here? Does it really matter if we get them?).

### DIFF
--- a/modules/contracts/src.ts/services/ethService.ts
+++ b/modules/contracts/src.ts/services/ethService.ts
@@ -666,31 +666,36 @@ export class EthereumChainService extends EthereumChainReader implements IVector
 
     // An anon fn to get the tx receipt, as we may require multiple retries with raised gas price.
     const getTransactionReceipt = async (): Promise<TransactionReceipt | undefined> => {
-      try {
-        return await ethers.provider.getTransactionReceipt(response.hash);
-      } catch (e) {
-        this.log.error({ method: "sendTxAndParseResponse", error: jsonifyError(e) }, "Transaction reverted");
-        await this.store.saveTransactionFailure(channelAddress, response.hash, e.message);
-        this.evts[ChainServiceEvents.TRANSACTION_FAILED].post({
-          error: e,
-          channelAddress,
-          reason,
-        });
-      }
-      return undefined;
+      // TODO: This should be replaced with a polling method (?)
+      // TODO: If we're polling here, we shouldn't be awaiting a receipt, but checking to see if it's available
+      // (i.e. it's been confirmed / mined).
+      return await ethers.provider.getTransactionReceipt(response.hash);
+      // TODO: If the tx has not yet been mined, return undefined.
+      // return undefined;
     }
 
-    // Poll for receipt.
-    let receipt: TransactionReceipt | undefined = await getTransactionReceipt();
-    // TODO: Should we have a diff. system of defaultRetries for this operation?
-    // NOTE: This loop won't execute if receipt is valid (not undefined).
-    for (let attempt = 1; attempt < this.defaultRetries && !receipt; attempt++) {
-      // Pause for 2 sec.
-      await delay(2000);
+    let receipt: TransactionReceipt | undefined;
+    try {
+      // Poll for receipt.
       receipt = await getTransactionReceipt();
-      if (receipt) {
-        break;
+      // TODO: Should we have a diff. system of defaultRetries for this operation?
+      // NOTE: This loop won't execute if receipt is valid (not undefined).
+      for (let attempt = 1; attempt < this.defaultRetries && !receipt; attempt++) {
+        // Pause for 2 sec.
+        await delay(2000);
+        receipt = await getTransactionReceipt();
+        if (receipt) {
+          break;
+        }
       }
+    } catch (e) {
+      this.log.error({ method: "sendTxAndParseResponse", error: jsonifyError(e) }, "Transaction reverted");
+      await this.store.saveTransactionFailure(channelAddress, response.hash, e.message);
+      this.evts[ChainServiceEvents.TRANSACTION_FAILED].post({
+        error: e,
+        channelAddress,
+        reason,
+      });
     }
 
     if (!receipt) {
--- a/modules/contracts/src.ts/services/ethService.ts
+++ b/modules/contracts/src.ts/services/ethService.ts
@@ -145,6 +145,7 @@ export class EthereumChainService extends EthereumChainReader implements IVector
       if (price.isError) {
         throw price.getError()!;
       }
+      // TODO: Scale up gas by percentage as specified by GAS_BUMP_PERCENT.
       const current = price.getValue().add(parseUnits("20", "gwei"));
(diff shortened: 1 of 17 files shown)


modules/contracts/src.ts/services/ethService.ts
```diff
@@ -684,6 +660,80 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
+      // return undefined;
+    }
+
+    let receipt: TransactionReceipt | undefined;
+    try {
+      // Poll for receipt.
+      receipt = await getTransactionReceipt();
+      // TODO: Should we have a diff. system of defaultRetries for this operation?
+      // NOTE: This loop won't execute if receipt is valid (not undefined).
+      for (let attempt = 1; attempt < this.defaultRetries && !receipt; attempt++) {
```
rhlsthrm, line 687: Should not be related to `defaultRetries` which is related to tx sending. I think this basically needs to be an infinite loop that keeps bumping gas. Maybe the ultimate timeout is when it gets to the MAX_GAS_PRICE we configure.

rhlsthrm (commented):
Looks good so far.

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -538,12 +543,12 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
     // should return undefined IFF tx didnt send based on validation in
     // fn
-    txFn: () => Promise<undefined | TransactionResponse>,
+    txFn: (gasPrice: BigNumber) => Promise<undefined | TransactionResponse>,
   ): Promise<Result<TransactionResponseWithResult | undefined, ChainError>> {
     const method = "sendTxWithRetries";
     const methodId = getRandomBytes32();
     const errors = [];
-    for (let attempt = 1; attempt++; attempt < this.defaultRetries) {
+    for (let attempt = 1; attempt < this.defaultRetries; attempt++) {
```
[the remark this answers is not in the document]
jakekidd: lol great point. this actually was not my code - it's funny, I was too confused by the mis-ordering of the statements I guess to notice that glaring issue.

I think this was copy pasted, and the original implementation had an "attempt" before the for loop

modules/contracts/src.ts/services/ethService.ts
jakekidd: Sounds good. TODO is left there, so safe to say I will handle that.

I'll go with a `while` loop and use gas price and !receipt to handle this

modules/contracts/src.ts/tests/integration/ethService.spec.ts
```diff
@@ -162,4 +159,8 @@ describe("EthereumChainService", function () {
     const res = await bobChainService.sendDefundTransferTx(transferState);
     expect(res.getValue()).to.be.ok;
   });
+
+  describe("revitalizeTxs", () => {
+    it("should start monitoring active txs", async () => {});
+  });
```
LayneHaber, line 165: What is this? Can we skip it at least? Comments would also be good

modules/server-node/src/services/store.ts
```diff
@@ -295,6 +295,48 @@ export class PrismaStore implements IServerNodeStore {
...
+        transactionHash: {
+          not: undefined
+        },
+        // TODO: Any other key fields that we know HAVE to be defined for receipts that we should include here?
+        // If these key receipt fields are undefined, then we know we never got
+        // a valid receipt for this transaction.
+        blockHash: undefined,
+        gasUsed: undefined,
+      }
+    });
```
LayneHaber, line 312: we have a `status` field that is `submitted` when it's submitted, and `mined` or `failed` otherwise. The only race around this status is when it was fired off, but not saved in the db. Maybe we need to save the `MinimalTransaction` under a different status before submitting to be extra careful? 

however, the status should be used as it is more clear.

modules/server-node/src/services/store.ts
```diff
@@ -295,6 +295,48 @@ export class PrismaStore implements IServerNodeStore {
...
+      blockNumber: t.blockNumber ?? undefined,
+      logs: t.logs ?? undefined,
+      contractAddress: t.contractAddress ?? undefined,
+      transactionIndex: t.transactionIndex ?? undefined,
+      root: t.root ?? undefined,
+      gasUsed: t.gasUsed ?? undefined,
+      logsBloom: t.logsBloom ?? undefined,
+      cumulativeGasUsed: t.cumulativeGasUsed ?? undefined,
+      byzantium: t.byzantium ?? undefined,
+    } as StoredTransaction))
```
LayneHaber, line 337: we should really add an entity conversion function for this

modules/contracts/src.ts/tests/integration/ethService.spec.ts
rhlsthrm: Whoa that should not be there.

modules/server-node/src/services/store.ts
rhlsthrm: If it's fired off but not saved in the db we don't save it and I don't think we should tbh.

modules/server-node/src/services/store.ts
rhlsthrm: Agreed, @jakekidd can you add this?

PR #599, opened 2021-04-30
new: bump 30% gas price.

sanchaymittal:


modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -497,6 +497,7 @@ export class EthereumChainReader implements IVectorChainReader {
       if (!gasPrice) {
         try {
           gasPrice = await provider.getGasPrice();
+          gasPrice = gasPrice.add(gasPrice.mul(30).div(100))
```
rhlsthrm, line 500: Move this to the top to be a declared constant.

rhlsthrm (changes requested):

modules/contracts/src.ts/services/ethReader.ts
rhlsthrm: By "this" I mean the `30`.

jakekidd: Should set constant to `0.3`.

modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -40,6 +40,7 @@ import { ChannelFactory, ChannelMastercopy, TransferDefinition, TransferRegistry
 import { Evt } from "evt";
 
 export const MinGasPrice = parseUnits("5", "gwei");
+export const BumpGasPrice = 30; // 30% bump gas price
```
rhlsthrm, line 43: Let's use the normal convention we use for these constants which would be `MIN_GAS_PRICE` and `BUMP_GAS_PRICE`.

rhlsthrm (changes requested):

modules/contracts/src.ts/services/ethReader.ts
sanchaymittal: gotchha

rhlsthrm (approved):

rhlsthrm (approved):

PR #601, opened 2021-05-01
Tx service robust

rhlsthrm:
~A couple issues I saw:
~- The `sendTxAndParseResponse` function should return the response immediately after it's sent (to chain).~
~- The `completed` function should be what handles all the monitoring and gas price bumping.~

Edit: Rethought of the above. All the submitted vs mined discrepancy needs to be handled in this tx service. All the callers should only care about the receipt when it's mined.

- [x] Tx isn't being resent at a higher gas price?

### DIFF
--- a/modules/contracts/src.ts/services/ethService.ts
+++ b/modules/contracts/src.ts/services/ethService.ts
@@ -428,7 +428,6 @@ export class EthereumChainService extends EthereumChainReader implements IVector
       }
 
       const receipt = await queuedResponse.completed;
-      console.log("queuedResponse.completed receipt: ", receipt);
       if (receipt.status === 0) {
         return Result.fail(new ChainError(ChainError.reasons.TxReverted, { receipt, method, methodId }));
       }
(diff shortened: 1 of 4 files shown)


modules/contracts/src.ts/services/ethService.ts
```diff
@@ -280,137 +318,172 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
-        ...response,
-        completed: async (confirmations?: number) => {
-          try {
-            const receipt = await response.wait(confirmations);
-            if (receipt.status === 0) {
-              return Result.fail(new ChainError(ChainError.reasons.TxReverted, { receipt }));
+        // completed callback should always return a receipt and abstract the waiting and bumping of gas prices
+
+        // use a promise here so that the whole confirmation is not held up in the queue
+        let tryNumber = 0;
```
jakekidd, line 355: This iteration variable (?) is never changed

rhlsthrm: Damn it you're right.

PR #602, opened 2021-05-03
Fix race when setting up channel

That3Percent:
This fixes a race condition which could lead to corrupted channel state


LayneHaber (approved):
delay usage subtle indeed

PR #612, opened 2021-05-06
Schema changes

rhlsthrm:


modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -144,8 +144,7 @@ model Transfer {
   // has onchainTransactionId always if withdrawal
   // has OnchainTransaction relation IFF we submitted (and have tx record)
   onchainTransactionId String?
-  transactionHash      String?
-  onchainTransaction   OnchainTransaction? @relation(name: "Transaction", fields: [transactionHash], references: [transactionHash])
+  onchainTransaction   OnchainTransactionReceipt? @relation(fields: [onchainTransactionId], references: [transactionHash])
```
jakekidd, line 147: like this

jakekidd, line 147: change the references to be `[id]`

bc the field `onchainTransactionId` should *reference* the field `id` on the related onchainTransaction

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
...
+  reason         String // no enums
+  error          String?
+  channelAddress String
+  channel        Channel @relation(fields: [channelAddress], references: [channelAddress])
+
+  createdAt DateTime @default(now())
+
+  confirmedReceipt OnchainTransactionReceipt? @relation(fields: [confirmedReceiptHash], references: [transactionHash])
+  confirmedReceiptHash String?
+  onchainTransactionAttempts OnchainTransactionAttempt[] @relation
```
jakekidd, line 200: so the final gas price for this tx will be the gas price belonging to the last tx in this list?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
...
+
+  @@map(name: "onchain_transaction")
+}
+
+model OnchainTransactionAttempt {
   // Response fields
   transactionHash String @id
   to              String
```
jakekidd, line 208: Shouldn't `to` go into the `OnchainTransaction` ?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -195,6 +214,18 @@ model OnchainTransaction {
   gasLimit        String
   gasPrice        String
 
+  createdAt DateTime @default(now())
+
+  onchainTransaction   OnchainTransaction? @relation(fields: [onchainTransactionId], references: [id])
+  onchainTransactionId String
```
jakekidd, line 220: Imho, I'm not a fan of this reverse relationship here. For the sake of minimization, can we keep it out?

jakekidd: I don't think we should ever retrieve 'individual' `onchainTransactionAttempt`

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
 }
 
 model OnchainTransaction {
+  // Response fields
+  id String @id @default(uuid())
+
+  // Channel fields
+  status         String  @default("submitted") // no enums
+  reason         String // no enums
+  error          String?
```
jakekidd, line 192: Does it make sense to have just 1 error here? Should we not record errors for each transaction attempt ? Or

jakekidd: Maybe 1 overall error here and an error column in attempt?

modules/server-node/prisma-sqlite/schema.prisma
jakekidd: Is there ever a time where the `to` will change across tx attempts?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -208,19 +239,15 @@ model OnchainTransaction {
...
-  channelAddress String
-  channel        Channel @relation(fields: [channelAddress], references: [channelAddress])
+  status            Int?
 
   createdAt DateTime @default(now())
 
-  @@unique([transactionHash, chainId])
-  @@map(name: "onchain_transaction")
-  Transfer Transfer[] @relation("Transaction")
+  transfer Transfer? @relation
```
jakekidd, line 246: This is going to create a really loopy relationship.

Imho, this should go from parent -> child -> grandchild

Transfer 1 -> 1 OnchainTransaction 1 -> 2 OnchainTransactionAttempt[]
Transfer 1 -> 1 OnchainTransaction 1 -> 2 OnchainTransactionReceipt[]

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -208,19 +239,15 @@ model OnchainTransaction {
...
+
+  onchainTransaction OnchainTransaction? @relation()
+  attempt OnchainTransactionAttempt? @relation
```
jakekidd, line 249: Same here as above. I think remove this reference here.

modules/server-node/prisma-sqlite/schema.prisma
jakekidd: Also - I believe `@relation` requires a method call structure and needs a `references` argument.

So... `@relation(name: "OnchainTransactionAttempt", references: [id]` ?

jakekidd: Ignore above

Remove `@relation` here, should only be on the many side of the relationship

jakekidd, line 200: Remove `@relation` here, should only be on the many side of the relationship

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -195,6 +214,18 @@ model OnchainTransaction {
   gasLimit        String
   gasPrice        String
 
+  createdAt DateTime @default(now())
+
+  onchainTransaction   OnchainTransaction? @relation(fields: [onchainTransactionId], references: [id])
+  onchainTransactionId String
+
+  receipt OnchainTransactionReceipt? @relation(fields: [transactionHash], references: [transactionHash])
```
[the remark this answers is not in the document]
jakekidd: It will be implied that the last attempt made in the parent `OnchainTransaction` was what succeeded.

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -195,6 +214,18 @@ model OnchainTransaction {
   gasLimit        String
   gasPrice        String
 
+  createdAt DateTime @default(now())
+
+  onchainTransaction   OnchainTransaction? @relation(fields: [onchainTransactionId], references: [id])
```
jakekidd, line 219: Remove optional `?` - this foreign key should absolutely be required

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
...
+  createdAt DateTime @default(now())
+
+  confirmedReceipt OnchainTransactionReceipt? @relation(fields: [confirmedReceiptHash], references: [transactionHash])
+  confirmedReceiptHash String?
+  onchainTransactionAttempts OnchainTransactionAttempt[] @relation
+
+  @@map(name: "onchain_transaction")
+}
+
+model OnchainTransactionAttempt {
```
jakekidd, line 205: Alternatively we could just call this `Attempt` and the one below `Receipt`

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -195,6 +214,18 @@ model OnchainTransaction {
...
+  onchainTransaction   OnchainTransaction? @relation(fields: [onchainTransactionId], references: [id])
+  onchainTransactionId String
+
+  receipt OnchainTransactionReceipt? @relation(fields: [transactionHash], references: [transactionHash])
+  @@map(name: "onchain_transaction_attempt")
+}
+
+model OnchainTransactionReceipt {
+  // Response fields
+  transactionHash   String   @id
```
jakekidd, line 228: So, one could make the argument that this hash should be omitted here as it is redundant. The receipt should apply to the *last* tx attempt in the tx attempt list column of the parent `OnchainTransaction`.

But gut feeling says we should keep this here for simplicity's sake

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
...
+  id String @id @default(uuid())
+
+  // Channel fields
+  status         String  @default("submitted") // no enums
+  reason         String // no enums
+  error          String?
+  channelAddress String
+  channel        Channel @relation(fields: [channelAddress], references: [channelAddress])
+
+  createdAt DateTime @default(now())
```
jakekidd, line 196: So, noticed that this table has `createdAt`, OnchainTransactionAttempt has `createdAt`, and OnchainTransactionReceipt has `timestamp` (which is a string). Could we maybe just use one format for all of these? Either a timestamp string type or createdAt Date type?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -195,6 +214,18 @@ model OnchainTransaction {
...
+
+  receipt OnchainTransactionReceipt? @relation(fields: [transactionHash], references: [transactionHash])
+  @@map(name: "onchain_transaction_attempt")
+}
+
+model OnchainTransactionReceipt {
+  // Response fields
+  transactionHash   String   @id
   // Receipt fields
   timestamp         String?
```
jakekidd, line 230: why is this optional? shouldn't all receipts have a timestamp?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -208,19 +239,15 @@ model OnchainTransaction {
   logs              String?
   cumulativeGasUsed String?
```
jakekidd, line 240: Imo, this ideally would belong to parent `OnchainTransaction`

modules/server-node/prisma-sqlite/schema.prisma
jakekidd: Nvm, ignore

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -208,19 +239,15 @@ model OnchainTransaction {
...
   byzantium         Boolean?
-
-  // Channel fields
-  status         String  @default("submitted") // no enums
-  reason         String // no enums
-  error          String?
-  channelAddress String
-  channel        Channel @relation(fields: [channelAddress], references: [channelAddress])
+  status            Int?
```
jakekidd, line 242: Shouldn't we remove status as it will be in the parent?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -208,19 +239,15 @@ model OnchainTransaction {
...
 
   createdAt DateTime @default(now())
 
-  @@unique([transactionHash, chainId])
-  @@map(name: "onchain_transaction")
-  Transfer Transfer[] @relation("Transaction")
+  transfer Transfer? @relation
+
+  onchainTransaction OnchainTransaction? @relation()
```
jakekidd, line 248: Remove optional `?`

Also, since f-key will be here, should be: `@relation(name: "OnchainTransaction", fields: [onchainTransactionId], references: [id])`

Also add a `onchainTransactionId` f-key field ^

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -208,19 +239,15 @@ model OnchainTransaction {
   logs              String?
   cumulativeGasUsed String?
   byzantium         Boolean?
-
-  // Channel fields
-  status         String  @default("submitted") // no enums
-  reason         String // no enums
-  error          String?
```
jakekidd, line 215: Should error not be present? Is that relegated to the `OnchainTransaction` parent?

modules/server-node/prisma-sqlite/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
...
+  status         String  @default("submitted") // no enums
+  reason         String // no enums
+  error          String?
+  channelAddress String
+  channel        Channel @relation(fields: [channelAddress], references: [channelAddress])
+
+  createdAt DateTime @default(now())
+
+  confirmedReceipt OnchainTransactionReceipt? @relation(fields: [confirmedReceiptHash], references: [transactionHash])
+  confirmedReceiptHash String?
```
jakekidd, line 199: nit: imho would prefer `confirmedTransactionHash`

modules/server-node/prisma-sqlite/schema.prisma
jakekidd: Ah, wait, this is the int status. The one where we check if `receipt.status === 0`

modules/server-node/prisma-postgres/schema.prisma
```diff
@@ -184,6 +183,26 @@ model EventSubscription {
 }
 
 model OnchainTransaction {
+  // Response fields
```
jakekidd, line 186: Please see comments made on the browser-node schema!

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -107,14 +108,15 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
     method: string,
     methodId: string,
     channelAddress: string,
     reason: TransactionReason,
     response: TransactionResponse,
-  ) {
+  ): Promise<void> {
     this.log.info({ method, methodId, channelAddress, reason, response }, "Tx submitted.");
-    await this.store.saveTransactionResponse(channelAddress, reason, response);
+    const res = await this.store.saveTransactionAttempt(onchainTransactionId, channelAddress, reason, response);
```
jakekidd, line 119: remove `const res`, we no longer return anything

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -244,24 +241,22 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
+        receipt = await provider.send("eth_getTransactionReceipt", [txHash]);
       }
       return Result.ok({ response, receipt });
     } catch (e) {
-      return Result.fail(
-        new ChainError(ChainError.reasons.TxNotFound, { error: e, transactionHash: tx.transactionHash }),
-      );
+      return Result.fail(new ChainError(ChainError.reasons.TxNotFound, { error: e, transactionHash: txHash }));
     }
   }
```
jakekidd, line 262: If we are removing speed up tx RPC method, lets remove this one as well? @rhlsthrm

modules/server-node/src/index.ts
```diff
@@ -1340,59 +1340,60 @@ server.post<{ Body: NodeParams.SendExitChannelTx }>(
...
+//     data: record.data,
+//     value: record.value,
+//     nonce: record.nonce,
+//     transactionHash: record.transactionHash,
+//   });
+//   if (result.isError) {
+//     return reply.status(500).send(jsonifyError(result.getError()!));
+//   }
+//   return reply.status(200).send({ transactionHash: result.getValue().transactionHash });
+// });
```
jakekidd, line 1396: I was thinking the same thing about speed up - imo, makes more sense to obfuscate / deprecate this functionality for now, relying instead on our own internal 'speeding up' logic. There may be individual use cases down the road where someone might like to use the speed up functionality, so we'll leave room to re-implement, but I agree with this move.

Left a note on ethService - do we want to remove that method for `speedUpTx` as well?

jakekidd (approved):

PR #624, opened 2021-05-19
Fix transaction restarts

LayneHaber:
## The Problem

Fixes issue where the A in ACID for prisma actually gets more of a D


LayneHaber:
Still TODO:
- potentially fix tests/method
- fix restore saving call

rhlsthrm (approved):
Looks good as a stopgap but we should probably move to stored procs.

PR #625, opened 2021-05-19
adjust logic for handling submit

jakekidd:

### DIFF
--- a/modules/contracts/src.ts/services/ethService.ts
+++ b/modules/contracts/src.ts/services/ethService.ts
@@ -507,7 +507,7 @@ export class EthereumChainService extends EthereumChainReader implements IVector
    * a timeout. If within the timeout there are *not* 10 confirmations,
    * the tx will be resubmitted at the same nonce.
    */
-   public async waitForConfirmation(chainId: number, responses: TransactionResponse[]): Promise<TransactionReceipt> {
+  public async waitForConfirmation(chainId: number, responses: TransactionResponse[]): Promise<TransactionReceipt> {
     const provider: JsonRpcProvider = this.chainProviders[chainId];
     if (!provider) {
       throw new ChainError(ChainError.reasons.ProviderNotFound);
--- a/modules/contracts/src.ts/services/ethService.ts
+++ b/modules/contracts/src.ts/services/ethService.ts
@@ -339,8 +339,13 @@ export class EthereumChainService extends EthereumChainReader implements IVector
     if (!signer) {
       return Result.fail(new ChainError(ChainError.reasons.SignerNotFound));
     }
+    // Used to track the number of attempts, regardless of whether a tx was successfully submitted
+    // on each.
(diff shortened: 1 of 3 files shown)


modules/contracts/src.ts/services/ethService.ts
```diff
@@ -368,53 +373,72 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
           } else {
-            response = tempResponse;
+            throw error;
           }
-
-          // NOTE: Response MUST be defined here because if it was NEVER defined (i.e. undefined on first iteration),
-          // we would have returned in prev block, and if it was undefined on this iteration we would not overwrite
-          // that value.
-          // Tx was submitted: handle saving to store.
-          await this.handleTxSubmit(onchainTransactionId, method, methodId, channelAddress, reason, response!);
```
jakekidd, line 407: So, working theory is that this line and the `response = tempResponse` line above were the problem lines.

Somehow, there was a response of `undefined` in a **subsuquent ** (2nd or greater number attempt) and we were *not* saving `tempResponse` to `response`... so we were calling `this.handleTxSubmit` and passing in a **stale response from the previous run**.

This resulted in uncaught prisma error - the txHash must be unique

jakekidd: As to why we got `undefined` in a subsequent response, I don't know. @LayneHaber or @rhlsthrm maybe you can help me out with that. @LayneHaber I think there was some reason for it... maybe re-explain. This might be worth writing a unit test for that edge case

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -368,53 +373,77 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
-          // passed into txFn to bail out at the time of sending.
-          let tempResponse = result.getValue();
-          if (!tempResponse) {
-            this.log.warn({ method, methodId, channelAddress, reason }, "Did not attempt tx");
-            // Iff this is the only iteration, then we want to go ahead return w/o saving anything.
-            if (tryNumber === 1) {
-              return Result.ok(undefined);
+            // We must check for confirmation in all previous transactions. Although it's most likely
+            // that it's the previous one, any of them could have been confirmed.
+            for (let i = 0; i < responses.length; i++) {
```
jakekidd, line 428: iter backwards

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -368,53 +373,77 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
+              try {
+                receipt = await this.waitForConfirmation(chainId, responses[i]);
+              } catch {}
```
jakekidd, line 431: catch error, check if timeout -should be timeout or something?

LayneHaber (approved):

PR #633, opened 2021-05-24
Load testing

LayneHaber:
## The Problem

Add the following load tests:
- How many transfers can `n` bots execute in a given time period?
- How many transfers can a bot create in channel in a given time period?
- How many bots can create `n` payments simultaneously up to some ceiling of active agents?


modules/auth/ops/entry.sh
```diff
@@ -51,7 +51,7 @@ then
 
 else
   echo "Starting node in dev-mode"
-  exec ./node_modules/.bin/nodemon \
+  exec nodemon \
```
rhlsthrm: Not liking this change at all. Is this necessary?

rhlsthrm (approved):

modules/auth/ops/entry.sh
LayneHaber: no

PR #639, opened 2021-05-27
Auction

rhlsthrm:
TODO:
- [ ] Run in test env.


viken33:
Changes on fee service and auction listener look great, I tried to test it locally but contracts test are failing, not sure why.
Let me know if I can help with anything.

PR #640, opened 2021-05-27
Implement custom RPC Provider with Fallback

jakekidd:

### DIFF
--- a/modules/types/src/network.ts
+++ b/modules/types/src/network.ts
@@ -1,4 +1,21 @@
-import { JsonRpcProvider } from "@ethersproject/providers";
+import { FilterByBlockHash, BlockWithTransactions, TransactionRequest } from "@ethersproject/abstract-provider";
+import { BigNumber, BigNumberish } from "@ethersproject/bignumber";
+import {
+  JsonRpcProvider,
+  FallbackProvider,
+  Block,
+  BlockTag,
+  EventType,
+  Filter,
+  Listener,
+  Log,
+  Network,
+  Provider,
+  Resolver,
+  TransactionReceipt,
+  TransactionResponse,
+} from "@ethersproject/providers";
+import { Deferrable } from "@ethersproject/properties";
 
 export type ChainProviders = {
   [chainId: number]: string;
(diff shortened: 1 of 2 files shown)


modules/types/src/network.ts
```diff
@@ -1,9 +1,261 @@
...
+
+  RPC_TIMEOUT: number = 10_000;
+  _isProvider: boolean = true;
+  _networkPromise: Promise<Network>;
+  _network: Network;
+  anyNetwork: boolean = false;
+
+  constructor(chainId: number, providers: string[] | JsonRpcProvider[], stallTimeout?: number) {
+    // We'll collect all the provider URLs as we hydrate each provider.
+    var providerUrls: string[] = [];
```
rhlsthrm, line 45: Ew `var`. `const` or `let` pls!

rhlsthrm (commented):
Need to do the `StaticJsonRpcProvider` instead of `JsonRpcProvider`.

modules/types/src/network.ts
jakekidd: hahaha
yeah old code but u got it

modules/types/src/network.ts
```diff
@@ -1,9 +1,253 @@
...
+    this.providerUrls = providerUrls;
+  }
+
+  send(method: string, params: any[]): Promise<any> {
+    if (this._provider instanceof JsonRpcProvider) {
+      return (this._provider as JsonRpcProvider).send(method, params);
+    } else {
+      const providers = (this._provider as FallbackProvider).providerConfigs.map(p => p.provider as JsonRpcProvider);
+      var errors: Error[] = [];
+      return Promise.race<any>(
```
jakekidd, line 92: Note that here is where we are executing the `Promise.race`. Essentially, in any other method call in this class, we literally just call the underlying provider. Fallback provider does this sort of 'race-like' operation we see here internally, under the hood.

However, there is no `FallbackProvider.send` method, so I've taken the liberty of implementing this myself. Assuming we had multiple provider URLs passed into constructor (and therefore we're using `FallbackProvider`), we're basically executing the RPC `send` call on every child provider (AS `JsonRpcProvider`) belonging to our `FallbackProvider`, and using a Promise.race to return the fastest one.

A delay for timeout of 10s rejects the Promise. Additionally, if all `send` calls error out, we reject the Promise with a list of errors.

jakekidd: Going to add the above in descriptive comments here.

LayneHaber:
# Update June 7

Work on this is deprioritized temporarily (will likely resume next week).

Currently blocked by an issue with the `ethprovider` image. Specifically, when running integration tests the contracts fail to deploy because the image does not recognize the symlinked types package.

When creating the `ethprovider` image, there are dependencies installed that are listed [here](https://github.com/connext/vector/blob/fallback-provider/modules/contracts/ops/package.json) which the image has access too. The symlinked packages are generally only accessible by the `vector_builder` image, which is the image used by the `server-node` package in the `duet` and `trio` stacks.

To move forward, we have to give the images used in the `chains` stack access to the symlinked packages. This means replacing the `ethprovider` image with the `builder` image within the `start-chains` script.

PR #641, opened 2021-05-28
Use safe block for read operations

LayneHaber:


modules/contracts/src.ts/services/ethReader.ts
```diff
@@ -843,6 +962,24 @@ export class EthereumChainReader implements IVectorChainReader {
     return Result.ok(cleaned);
   }
 
+  private async getSafeBlockNumber(chainId: number): Promise<Result<number, ChainError>> {
```
rhlsthrm, line 965: We should think about caching this in memory, it's gonna be crazy to constantly query this on every request.

rhlsthrm (approved):

modules/contracts/src.ts/services/ethReader.ts
LayneHaber: We could do add a block listener

rhlsthrm (approved):

PR #642, opened 2021-05-28
Add tx hash

rhlsthrm:

### DIFF
--- a/modules/server-node/src/services/store.ts
+++ b/modules/server-node/src/services/store.ts
@@ -550,6 +550,7 @@ export class PrismaStore implements IServerNodeStore {
         where: { transferId },
         data: {
           onchainTransaction: { connect: { id: record.id } },
+          transactionHash: withdrawCommitment.transactionHash,
         },
       });
     }


LayneHaber (approved):
wau

modules/server-node/src/services/store.ts
```diff
@@ -550,6 +550,7 @@ export class PrismaStore implements IServerNodeStore {
         where: { transferId },
         data: {
           onchainTransaction: { connect: { id: record.id } },
+          transactionHash: withdrawCommitment.transactionHash,
```
jakekidd, line 553: This needs to be the tx hash because this withdrawCommitment comes from the user, correct? Therefore there won't be an `OnchainTransaction` isntance with an `OnchainTransactionAttempt` child?

Bc normally I would say just refer to `onchainTransaction.attempts[i].transactionHash`, or `onchainTransaction.receipt.transactionHash`.

rhlsthrm: This can either come from the counterparty or from ourselves, so in either case, we need to save it. If it comes from the counterparty, there's no notion of an `onchainTransaction`.

rhlsthrm: This is covered by the new message we added `onReceiveWithdrawalSubmittedMessage`. The router sends a message to the counterparty indicating that it submitted (and confirmed) the tx.

PR #649, opened 2021-06-03
Track nonce internally

LayneHaber:
## The Problem

- Sometimes seeing error where nonce is being used twice for different transactions, resulting in `replacement fee too low` errors

## The Solution

- Track nonce within the `ethService` class

### DIFF
--- a/modules/contracts/src.ts/services/ethService.ts
+++ b/modules/contracts/src.ts/services/ethService.ts
@@ -108,17 +108,6 @@ export class EthereumChainService extends EthereumChainReader implements IVector
     return signer;
   }
 
-  private async updateNonce(chainId: number, usedNonce: number, signer: Signer): Promise<void> {
-    const pending = await signer.getTransactionCount("pending");
-    const stored = this.nonces.get(chainId) ?? 0;
-    const incremented = usedNonce + 1;
-    // Ensure the nonce you store is *always* the greatest of the values
-    if (stored >= incremented && stored >= pending) {
-      return;
-    }
-    this.nonces.set(chainId, incremented > pending ? incremented : pending);
-  }
-
   /// Upsert tx submission to store, fire tx submit event.
   private async handleTxSubmit(
     onchainTransactionId: string,
@@ -219,11 +208,20 @@ export class EthereumChainService extends EthereumChainReader implements IVector
     const task = async (): Promise<Result<TransactionResponse | undefined, Error>> => {
       try {
         // Send transaction using the passed in callback.
-        const nonceToUse: number = nonce ?? this.nonces.get(chainId) ?? (await signer.getTransactionCount("pending"));
+        const stored = this.nonces.get(chainId);
+        const nonceToUse: number = nonce ?? stored ?? (await signer.getTransactionCount("pending"));
         const response: TransactionResponse | undefined = await txFn(gasPrice, nonceToUse);
-        // After calling tx fn, set nonce to usedNonce + 1 IFF the
-        // nonce wasnt pre-specified (i.e. not re-sending tx)
-        await this.updateNonce(chainId, nonceToUse, signer);
+        // After calling tx fn, set nonce to the greatest of
+        // stored, pending, or incremented
+        const pending = await signer.getTransactionCount("pending");
+        const incremented = nonceToUse + 1;
+        // Ensure the nonce you store is *always* the greatest of the values
+        const toCompare = stored ?? 0;
+        if (toCompare >= incremented && toCompare >= pending) {
+          // return without updating
+          return Result.ok(response);
+        }
+        this.nonces.set(chainId, incremented > pending ? incremented : pending);
         return Result.ok(response);
       } catch (e) {
         return Result.fail(e);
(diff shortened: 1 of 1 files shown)


modules/contracts/src.ts/services/ethService.ts
```diff
@@ -195,15 +196,22 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
       async (): Promise<Result<TransactionResponse | undefined, Error>> => {
         try {
           // Send transaction using the passed in callback.
-          const actualNonce: number = nonce ?? (await signer.getTransactionCount("pending"));
+          const actualNonce: number =
+            nonce ?? this.nonces.get(chainId) ?? (await signer.getTransactionCount("pending"));
           const response: TransactionResponse | undefined = await txFn(gasPrice, actualNonce);
+          // After calling tx fn, set nonce to usedNonce + 1 IFF the
+          // nonce wasnt pre-specified (i.e. not re-sending tx)
+          if (typeof nonce === undefined) {
```
jakekidd, line 212: why `typeof` ? Can't we just say `if (nonce === undefined)` ?

LayneHaber: that would ignore the `null` case

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -195,21 +208,29 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
-          const actualNonce: number = nonce ?? (await signer.getTransactionCount("pending"));
-          const response: TransactionResponse | undefined = await txFn(gasPrice, actualNonce);
-          return Result.ok(response);
-        } catch (e) {
-          return Result.fail(e);
-        }
-      },
-    );
+    if (!this.queues.has(chainId)) {
+      return Result.fail(new ChainError(ChainError.reasons.SignerNotFound));
```
jakekidd, line 216: interesting...

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -195,21 +208,29 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
+        const response: TransactionResponse | undefined = await txFn(gasPrice, nonceToUse);
+        // After calling tx fn, set nonce to usedNonce + 1 IFF the
+        // nonce wasnt pre-specified (i.e. not re-sending tx)
+        await this.updateNonce(chainId, nonceToUse, signer);
+        return Result.ok(response);
+      } catch (e) {
+        return Result.fail(e);
+      }
+    };
+    const result = await this.queues.get(chainId)!.add(task);
```
jakekidd, line 232: woah 🤯

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -106,6 +108,17 @@ export class EthereumChainService extends EthereumChainReader implements IVector
     return signer;
   }
 
+  private async updateNonce(chainId: number, usedNonce: number, signer: Signer): Promise<void> {
```
jakekidd, line 111: again, only nitpick I have here is that this should only ever be called within the queue, so I don't like this code existing outside of it.

if someone called from elsewhere, it would be disaster. Obvi no one will, but still... idk

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -195,21 +208,29 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
+    }
+    // Define task to send tx with proper nonce
+    const task = async (): Promise<Result<TransactionResponse | undefined, Error>> => {
+      try {
+        // Send transaction using the passed in callback.
+        const nonceToUse: number = nonce ?? this.nonces.get(chainId) ?? (await signer.getTransactionCount("pending"));
+        const response: TransactionResponse | undefined = await txFn(gasPrice, nonceToUse);
+        // After calling tx fn, set nonce to usedNonce + 1 IFF the
+        // nonce wasnt pre-specified (i.e. not re-sending tx)
+        await this.updateNonce(chainId, nonceToUse, signer);
```
jakekidd, line 226: here we pass in `nonceToUse`, but the argument is labeled `usedNonce`

maybe to be more loyal to that, we could actually use `response.nonce` here?

jakekidd: like, pass in the nonce that was actually used by the tx. I would imagine they would be the same, but this code likes to act UP

jakekidd (approved):

modules/contracts/src.ts/services/ethService.ts
```diff
@@ -195,21 +197,38 @@ export class EthereumChainService extends EthereumChainReader implements IVector
...
+        const nonceToUse: number = nonce ?? stored ?? (await signer.getTransactionCount("pending"));
+        const response: TransactionResponse | undefined = await txFn(gasPrice, nonceToUse);
+        // After calling tx fn, set nonce to the greatest of
+        // stored, pending, or incremented
+        const pending = await signer.getTransactionCount("pending");
+        const incremented = nonceToUse + 1;
+        // Ensure the nonce you store is *always* the greatest of the values
+        const toCompare = stored ?? 0;
+        if (toCompare >= incremented && toCompare >= pending) {
+          // return without updating
```
jakekidd, line 221: just to clean up - maybe we could reverse this if condition and put the `this.nonces.set` inside the if block

and then after the if block put the return statement

modules/contracts/src.ts/services/ethService.ts
LayneHaber: It's specified to the `txFn` so I don't really think that matters, but sure

modules/contracts/src.ts/services/ethService.ts
LayneHaber: Fixed

modules/contracts/src.ts/services/ethService.ts
LayneHaber: Safety check more than anything

